Issue #5, opened 2021-03-09
Change directory organisation to reflect upstream LLVM default sysroot path


vhscampos:
Upstream LLVM introduced a default sysroot path when targeting bare metal ([link](https://reviews.llvm.org/rG275592e7141)). It follows this pattern:

`DRIVER_PATH/lib/clang-runtimes/TARGET_TRIPLE`

Currently we install newlib and compiler-rt to `DRIVER_PATH/targets/TARGET_TRIPLE`, and this is used as sysroot.

The proposal here is to align our sysroot with upstream LLVM. Hence, we'd install newlib and compiler-rt to the new default path for bare metal compilations. Consequently, occurrences of `--sysroot` in configuration files can be removed as the default can be used.

vhscampos:
Fixed in #7

PR #7, opened 2021-03-10
Change sysroot to DRIVER_PATH/lib/clang-runtimes

vhscampos:
Upstream LLVM was changed in 275592e714130345a481a5cb889c89b73a98614f to
specify a default path for sysroot when targeting bare metal.

This default path is DRIVER_PATH/lib/clang-runtimes/TARGET.

In order to shorten our own modifications on top of LLVM, present in
patches/, this patch reorganises our installation directory's hierarchy.
Newlib and compilerrt now install to the default path, and --sysroot is
removed from our configuration files.

It would be ideal to remove our entire local patch altogether. However,
our configuration files still rely on our patch in order to specify
relative paths to crt0 and -L.


vhscampos:
Related to #5

statham-arm (approved):

PR #11, opened 2021-03-16
Python scripts overhaul

miyuki:
This series of patches refactors the Python scripts as follows:
- makes the scripts comply to pep8
- adds pylint configuration and makes fixes all pylint warnings
- fixes some other style issues.

### DIFF
--- a/scripts/repos.py
+++ b/scripts/repos.py
@@ -18,180 +18,183 @@
 #
 
 import argparse
-import yaml
 import os
 import re
-import git
 import textwrap
 import sys
 
-def die(msg, rv=1):
+import git
+import yaml
+
+def die(msg, ret_val=1):
     """Exit from program with the specified message and return value."""
     print('Error: {}'.format(msg))
-    sys.exit(rv)
+    sys.exit(ret_val)
 
 def warn(msg):
     """Print a warning message."""
     print('Warning: {}'.format(msg))
 
-class ModuleTC(object):
+class ModuleTC:
     """A building block of the LLVM Embedded Toolchain for Arm."""
 
-    def __init__(self, d):
-        assert isinstance(d, dict)
-        for k in ['Name', 'URL', 'Revision']:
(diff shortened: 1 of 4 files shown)


scripts/repos.py
```diff
@@ -210,78 +197,73 @@ def check_repositories_status(repositories, tc_version):
...
     else:
-        os.mkdir(repositories)
-
-    repos = list(tc_version.modules.keys())
+        os.mkdir(checkout_path)
 
-    print('Clone ({} @ {}):'.format(repositories, tc_version.revision))
-    for repo in repos:
+    print('Clone ({} @ {}):'.format(checkout_path, tc_version.revision))
+    for repo in tc_version.modules.keys():
```
statham-arm, line 219: This could be more compact still as
```for repo, module in tc_version.modules.items():```

File: scripts/repos.py [hunk 2]
```diff
@@ -115,61 +123,74 @@ def print_versions(versions, verbose):
...
             status[rel_path]['Branch'] = repo.active_branch.name
     return status
 
+
 def print_repositories_status(repositories):
-    """ Report the state of each git repository."""
-    assert isinstance(repositories, str), "Expecting a string for the directory name."
+    """Report the state of each git repository."""
+    assert isinstance(repositories,
+                      str), "Expecting a string for the directory name."
```
statham-arm: Urgh! I see that it gets the line length shorter, but I'm not sure it actually improves readability. How about
```
assert isinstance(repositories, str), (
    "Expecting a string for the directory name.")
```

scripts/repos.py
miyuki: Done

scripts/repos.py
miyuki: I've updated the patch series. I decided to leave the white space formatting part as-is because some lines (including most assert statements) are removed in the third patch. The third patch also deals with formatting issues left after the tool.

statham-arm (approved):

PR #12, opened 2021-03-18
support for v7 and v8 targets

stuij:
I've added a number of armv7 and armv8 target configurations to the buildscript. To account for more than one options, I've added an environment variable, `LIBRARY_SPEC`, that can be populated with the following values:

    armv8.1m.main_hard_fp_mve
    armv8.1m.main_hard_fp_nomve
    armv8.1m.main_soft_nofp_nomve
    armv8m.main_hard_fp
    armv8m.main_soft_nofp
    armv7em_hard_fpv4_sp_d16
    armv7em_hard_fpv5_d16
    armv7em_soft_nofp
    armv7m_soft_nofp
    armv6m_soft_nofp

hard and soft stand for fp linkage

They all build correctly with newlib and compiler-rt, and seem to do the right thing with regards to compiler-rt floating-point files to choose from: do we support double float, just single float, or neither of these.

At least two things I'd like input on:
- are we happy with these options for v7 and v8
- are we happy with the naming of these options

I've roughly based the options on which explicit cores LLVM supports.

Currently the generated files for these are dispatched to folders that are only differentiated by arch name. So if we build armv8.1m.main with soft and hard fp in succession, the one will overwrite the other. This week we removed being able to specify non-default folders for newlib and compiler-rt and the clang config file, so I thought not to put extra effort into this. I guess this will be resolved when properly implementing multilib support.

### DIFF
--- a/build-newlib.sh
+++ b/build-newlib.sh
@@ -20,8 +20,14 @@
 build_newlib () {
   pushToCleanDir $BUILD_DIR/newlib
 
+  NEWLIB_HW_FP="--disable-newlib-hw-fp"
+  if [ $NEWLIB_FP_SUPPORT = true ]; then
+      NEWLIB_HW_FP="--enable-newlib-hw-fp"
+  fi
+
   CC_FOR_TARGET="$TARGET_LLVM_PATH/bin/clang -target $TARGET -ffreestanding" \
   CXX_FOR_TARGET="$TARGET_LLVM_PATH/bin/clang++ -target $TARGET -ffreestanding" \
+  CFLAGS_FOR_TARGET="${FLAGS}" \
   AR_FOR_TARGET="$TARGET_LLVM_PATH/bin/llvm-ar" \
   NM_FOR_TARGET="$TARGET_LLVM_PATH/bin/llvm-nm" \
   AS_FOR_TARGET="$TARGET_LLVM_PATH/bin/llvm-as" \
@@ -37,7 +43,8 @@ build_newlib () {
     --disable-newlib-supplied-syscalls \
     --enable-newlib-io-c99-formats \
     --disable-nls \
-    --enable-lite-exit
+    --enable-lite-exit \
+    $NEWLIB_HW_FP
 
   make -j$NPROC
   make install
--- a/config.sh
+++ b/config.sh
@@ -28,12 +28,30 @@ readonly HOST_LLVM_PATH="${HOST_LLVM_PATH:-/usr/bin}"
 # How many CPUs to use for building the LLVM baremetal toolchain for Arm:
 readonly NPROC="${NPROC:-$(getconf _NPROCESSORS_ONLN)}"
 
-# Default triple of the toolchain:
-readonly TARGET=armv6m-none-eabi
+# The Arm sub-arch we want to use:
+readonly ARCH=${ARCH:-"armv6m"}
+
+# Architecture options:
+# (good overview: https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html)
+readonly ARCH_OPTIONS=${ARCH_OPTIONS:-""}
+
+# Float ABI
+readonly FLOAT_ABI=${FLOAT_ABI:-"soft"}
 
 #
 # No modification should be necessary beyond this point.
 #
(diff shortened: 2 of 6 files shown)


stuij:
I've made a script to test building all these variants, which might aid in testing:

    #!/usr/bin/bash
    
    set -e
    
    if [ $# -ne 0 ]; then
        ./setup.sh
        ./build.sh prepare
        USE_CCACHE=1 USE_NINJA=1 NPROC=10 ./build.sh clang
    fi
    
    declare -a specs=(
        "armv8.1m.main_hard_fp_mve"
        "armv8.1m.main_hard_fp_nomve"
        "armv8.1m.main_soft_nofp_nomve"
        "armv8m.main_hard_fp"
        "armv8m.main_soft_nofp"
        "armv7em_hard_fpv4_sp_d16"
        "armv7em_hard_fpv5_d16"
        "armv7em_soft_nofp"
        "armv7m_soft_nofp"
        "armv6m_soft_nofp"
    )
    
    declare -a steps=(
        newlib
        compilerrt
        configure
        package
    )
    
    for spec in "${specs[@]}"; do
        for step in "${steps[@]}"; do
            LIBRARY_SPEC=$spec USE_CCACHE=1 USE_NINJA=1 NPROC=10 ./build.sh $step
            build-llvm-toolchain $spec $step
        done
    done

vhscampos (changes requested):
Small typo in commit message: "concidered". #

vhscampos (commented):
I reckon "+nofp" is redundant when the float ABI is soft, isn't it?
Likewise, "+fp" seems to be redundant if hard FP ABI is used.

config.sh
```diff
@@ -28,12 +28,85 @@ readonly HOST_LLVM_PATH="${HOST_LLVM_PATH:-/usr/bin}"
 # How many CPUs to use for building the LLVM baremetal toolchain for Arm:
 readonly NPROC="${NPROC:-$(getconf _NPROCESSORS_ONLN)}"
 
-# Default triple of the toolchain:
-readonly TARGET=armv6m-none-eabi
+
+LIBRARY_SPEC="${LIBRARY_SPEC:-armv6m}"
```
voltur01, line 32: This is expected to be the default, however it is not in the list below _soft_nofp suffix is missing.

stuij:
> I reckon "+nofp" is redundant when the float ABI is soft, isn't it?
> Likewise, "+fp" seems to be redundant if hard FP ABI is used.

I prefer to be somewhat regular with the naming. Yes we could be more minimal, but this way the differences between the options are easier to reason about. Clang options are really non-obvious, and things like `fp` mean different things between architecture versions. On top of that hard dosn't necessarily mean fp. For 8.1m we still want fp abi hard because we need the fp registers which are used for mve integer mode.

Also this might make automation a bit easier, but that's not really a main goal.

stuij:
I've changed up the 8.1m options were a bit trickier to capture with the MVE options invovled. Especially MVE integer mode which needs the fp registers, but can be enabled without fp support. This needed some changes in newlib.

For reference the 8.1m options now cover the options in: https://developer.arm.com/documentation/101754/0616/armclang-Reference/Other-Compiler-specific-Features/Supported-architecture-feature-combinations-for-specific-processors

- armv8.1m.main_hard_fp: covers all the fp options with or without the two MVE modes
- armv8.1m.main_hard_nofp_mve: MVE integer mode without fp
- armv8.1m.main_soft_nofp_nomve: no MVE and no fp

build-newlib.sh
```diff
@@ -20,8 +20,14 @@
 build_newlib () {
   pushToCleanDir $BUILD_DIR/newlib
 
+  ENABLE_FP="--disable-newlib-hw-fp"
```
vhscampos, line 23: I think this could be better named. We could have a case where ENABLE_FP="--disable-..." which is counter intuitive.

Maybe simply NEWLIB_HW_FP?

stuij: yes, good idea. done.

config.sh
stuij: fixed

vhscampos (approved):

PR #13, opened 2021-03-31
Add special case for Clang in deciding floating-point endianness

vhscampos:
newlib's ieeefp.h determines floating-point endianness by checking the
\_\_VFP_FP__ macro. If this macro is defined, then newlib assumes that the
target architecture uses the VFP architecture for FPs. In this sort of
arch, IEEE FP values have their word-level endianness the same as their
byte-level endianness.

For example, a double-precision FP value of 1 would be represented in
memory like this (in little-endian ARM targets):

<- lowest addr    highest addr ->
1  0  0  0 | 0  0  0  0

Otherwise, i.e. if \_\_VFP_FP__ macro is undefined, then newlib assumes the
FPA architecture for FPs instead. In this case, IEEE FP values are
always big-endian at word level, while at the byte-level they follow the
target architecture's endianness.

For example, the same double-precision FP value of 1 would be
represented like this (in little-endian ARM targets):

<- lowest addr    highest addr ->
0  0  0  0 | 1  0  0  0

The problem is: \_\_VFP_FP__ cannot be used to determine that if using
Clang. In this compiler, the property:

\_\_VFP_FP__ is defined if and only if the VFP architecture is in use.

is false. That's because Clang only defines \_\_VFP_FP__ if an FPU is
enabled, which is not the case when, for instance, soft-float ABI is in
place.

This patch adds another preprocessor check to ieefp.h. Now, if the
compiler used is Clang, newlib can assume the endianness defined by VFP.

### DIFF
--- a/patches/newlib-0.1.patch
+++ b/patches/newlib-0.1.patch
@@ -300,3 +300,23 @@ index 64e4aeb9a..c0e5e9049 100644
  #include <math.h>
  
  float
+diff --git a/newlib/libc/include/machine/ieeefp.h b/newlib/libc/include/machine/ieeefp.h
+index aa8a1903b..19be99ecc 100644
+--- a/newlib/libc/include/machine/ieeefp.h
++++ b/newlib/libc/include/machine/ieeefp.h
+@@ -72,7 +72,14 @@
+    byte ordering was big or little endian depending upon the target.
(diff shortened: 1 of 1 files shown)


miyuki:
LGTM

miyuki (approved):

PR #15, opened 2021-04-01
Port build scripts from bash to Python

miyuki:
This patch ports all functionality implemented as shell scripts to
Python with the following exceptions:
- build-misc.sh is not ported because it is unused
- source tarballs are not built because they need some rework: the
  Python scripts support out-of-tree build, whereas the current source
  tarball generation method relies on LLVM and newlib repositories
  being checked out in the main source tree
- lldb is not built because it is not being installed

The new scripts add several important new features:
- Control over locations of all directories related to the build
  process: source, external repositories, build directory, install
  directory, target directory for tarballs (out-of-tree build)
- Incremental builds
- Ability to specify a list of targets to build libraries for
- Two levels of logging verbosity

The existing shell scripts and the new Python scripts give identical
results (modulo whitespace differences in .cfg files) when used with
equivalent sets of command line options and environment variables.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 .*.swp
+.idea
 repos*/
 build*/
 venv/
 *LLVMEmbeddedToolchainForArm*
+__pycache__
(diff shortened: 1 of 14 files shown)


amilendra (approved):
LGTM.

amilendra (approved):
LGTM.

amilendra (approved):

PR #16, opened 2021-04-07
Miscellaneous bug fixes in Python build scripts

miyuki:
1. Fix handling of --variants all (the default value)
2. Correctly parse Clang versions which include git tag, e.g.
   "clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)"
3. Use /usr/bin as the default host toolchain directory, this allows
   building on both Ubuntu 18.04 and 20.04 without having to specify
   the toolchain directory manually.


stuij (approved):
LGTM

PR #21, opened 2021-04-14
Prerequisite patches for cross-compilation support

miyuki:
The first patch adds support for host toolchain kind selection (and GCC as an example). The second patch (--build-mode) just makes life a bit easier when switching between host toolchains.

### DIFF
--- a/scripts/check.py
+++ b/scripts/check.py
@@ -16,7 +16,9 @@
 import functools
 import logging
 import os
+import re
 import shutil
+from typing import Optional
 import util
 
 import config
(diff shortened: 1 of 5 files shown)


scripts/check.py
```diff
@@ -133,7 +182,7 @@ def _check_tool(cfg: config.Config, bin_name: str, name: str,
 def check_prerequisites(cfg: config.Config) -> None:
     """Check availability and versions of all required prerequisite software."""
     is_ok = True
-    is_ok = is_ok and _check_host_compiler(cfg)
+    is_ok = is_ok and _check_toolchain(cfg, cfg.host_toolchain)
```
vhscampos, line 185: Since `host_toolchain` is a member of `cfg` anyway, consider taking `cfg` as the single argument in `_check_toolchain` and then retrieving `host_toolchain` inside it.

vhscampos (changes requested):

vhscampos (approved):
LGTM

scripts/check.py
miyuki: When cross compiling we will need two different toolchains, so I would prefer not to change this.

vhscampos: Ok, so LGTM

PR #23, opened 2021-04-19
Add support for Linux -> Windows cross-compilation

miyuki:
This patch adds support for cross-compiling the Arm toolchain using
Mingw-w64 (targeting x86-64 Windows).

LLVM build requires a native toolchain because it needs a native
tablegen binary during build. For this reason build.py now has
two additional options: --native-toolchain and --native-toolchain-dir.

Cross-build of LLVM only requires native tablegen and llvm-config.
Runtime libraries (compiler-rt and newlib) need a full native
toolchain to be built. Currently we always build full native LLVM when
performing cross-compilation. In the future we might want to only
build tablegen and llvm-config if the user did not request to build
the runtime libraries.

Native Windows build has not been implemented yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,9 @@
 .*.swp
 .idea
-repos*/
-build*/
-venv/
-*LLVMEmbeddedToolchainForArm*
+/repos*/
+/build*/
+/install*/
+/venv/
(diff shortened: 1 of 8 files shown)


statham-arm (approved):
LGTM!

PR #24, opened 2021-04-19
redid newlib patch for hardfp libm failures

stuij:
Previously, as suggested on the newlib website, we used the
`--enable-newlib-hw-fp` flag to toggle between libm directories and also used
this flag to not use Arm floating point instructions.

However it turns out that that alternate fp libm directory is part of an
experiment that has been abandoned. The right way to address this is to check
for the availability of floating point in the C file itself. Also because this
won't be disruptive for current uses of the library.

I also added a new llvm patch file for HEAD, as since we've merged a fix
upstream, 0.1 and HEAD, differ in what patch needs to be applied to them.

### DIFF
--- a/patches/newlib-0.1.patch
+++ b/patches/newlib-0.1.patch
@@ -1,28 +1,3 @@
-diff --git a/configure b/configure
-index 5db527014..05b82fbfc 100755
---- a/configure
-+++ b/configure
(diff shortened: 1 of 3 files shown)


miyuki:
Could you please split out the new LLVM patch into a separate commit?

stuij:
done. made a separate pull request: https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/pull/25

miyuki (approved):

PR #26, opened 2021-04-19
Fix the list of scripts in setup.py

miyuki:
The scripts list should only contain Python scripts that need to be
installed in the bin directory of the Python sandbox.

This fixes the build of Python sandbox (virual environment) with
Python 3.6.


stuij (approved):
LGTM

PR #28, opened 2021-04-22
add Newlib patch file for HEAD

stuij:
As some newlib changes we did for the LLVM Embedded toolchain have been merged
upstream, we need a separate Newlib patch file for HEAD.

### DIFF
--- a/patches/newlib-HEAD.patch
+++ b/patches/newlib-HEAD.patch
@@ -1,0 +1,108 @@
+diff --git a/libgloss/arm/crt0.S b/libgloss/arm/crt0.S
+index 8490bde2f..8b85b28f4 100644
+--- a/libgloss/arm/crt0.S
++++ b/libgloss/arm/crt0.S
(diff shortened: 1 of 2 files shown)


miyuki:
Could you please mention the upstream changes (https://reviews.llvm.org/D100372 and your newlib change) in the commit message?

stuij:
This patch is just related to my newlib change. I've updated the patch to add the Newlib change.

vhscampos:
It also removes the part of the diff which touches newlib/libc/include/machine/ieeefp.h
I suggest we mention in the commit msg that that particular piece of diff is not necessary anymore when we use LLVM HEAD.

stuij:
Ah, I accidentally removed that :D

Also added info on the llvm patch.

miyuki:
Also, please use the official newlib mirror: https://github.com/mirror/newlib-cygwin/

miyuki (approved):
LGTM

PR #30, opened 2021-04-26
Miscellaneous fixes

miyuki:
The first two patches fix the code base to pass more linters (flake8 which checks PEP 8 conformance and mypy which checks type annotations). The third one fixes the code samples build.

### DIFF
--- a/samples/Makefile.conf
+++ b/samples/Makefile.conf
@@ -18,4 +18,4 @@
 ifndef VERSION
     VERSION=0.1
 endif
-BIN_PATH=../../../LLVMEmbeddedToolchainForArm-${VERSION}/bin/
+BIN_PATH=../../../install-${VERSION}/LLVMEmbeddedToolchainForArm-${VERSION}/bin/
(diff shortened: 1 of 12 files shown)


scripts/check.py
```diff
@@ -160,16 +160,18 @@ def _check_availability(bin_name: str, name: str = None) -> bool:
     if shutil.which(bin_name) is None:
         if name is None:
             name = bin_name
-        print('{} not found.'.format(name))
+        logging.error('%s not found', name)
```
vhscampos, line 163: Was this (and other related changes) caught by mypy?
If not, please move them into a new patch.

samples/Makefile.conf
```diff
@@ -18,4 +18,4 @@
 ifndef VERSION
     VERSION=0.1
 endif
-BIN_PATH=../../../LLVMEmbeddedToolchainForArm-${VERSION}/bin/
+BIN_PATH=../../../install-${VERSION}/LLVMEmbeddedToolchainForArm-${VERSION}/bin/
```
vhscampos, line 21: If VERSION is HEAD, I reckon this won't work.

vhscampos (changes requested):

scripts/check.py
miyuki: No, it wasn't. Have removed the change.

vhscampos (approved):

PR #31, opened 2021-04-27
Rename compiler-rt libraries after installation

miyuki:
Since LLVM commit https://reviews.llvm.org/D98452 the Arm compiler-rt
library names no longer include the full architecture name
(e.g. armv6m), but rather just "arm" or "armhf". This does not work
correctly with our configuration of Clang (the driver uses the full
architecture name).

This patch adds a step which renames the installed compiler-rt
libraries so that their names match the names generated by Clang.

### DIFF
--- a/scripts/make.py
+++ b/scripts/make.py
@@ -15,6 +15,7 @@
 
 import logging
 import os
+import re
 from typing import Dict, List, Mapping, Tuple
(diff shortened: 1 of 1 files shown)


vhscampos (approved):
LGTM

PR #32, opened 2021-04-27
Add tests to the build process

vhscampos:
This patch adds a 'test' phase to the build system.

For now we have added one smoke test, which is the same as
baremetal-semihosting from the 'samples' directory.

If qemu-arm is present in PATH, tests are built and run. Otherwise,
they are only built.

### DIFF
--- a/scripts/config.py
+++ b/scripts/config.py
@@ -110,6 +110,7 @@ class Action(enum.Enum):
     COMPILER_RT = 'compiler-rt'
     LIBCXX = 'libcxx'
     CONFIGURE = 'configure'
+    TEST = 'test'
     PACKAGE = 'package'
     ALL = 'all'
 
--- a/scripts/make.py
+++ b/scripts/make.py
@@ -526,3 +526,27 @@ class ToolchainBuild:
             raise util.ToolchainBuildError from ex
         if cfg.is_cross_compiling:
             self._copy_runtime_to_native(lib_spec)
+
+    def _run_smoke_tests(self, _lib_spec: config.LibrarySpec) -> None:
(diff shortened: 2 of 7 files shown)


scripts/config.py
```diff
@@ -110,6 +110,7 @@ class Action(enum.Enum):
     COMPILER_RT = 'compiler-rt'
     LIBCXX = 'libcxx'
     CONFIGURE = 'configure'
+    TEST = 'test'
```
miyuki, line 113: TEST should not be part of ALL because that would break Windows cross-builds.

scripts/make.py
```diff
@@ -526,3 +526,27 @@ def compiler_str(bin_name: str) -> str:
             raise util.ToolchainBuildError from ex
         if cfg.is_cross_compiling:
             self._copy_runtime_to_native(lib_spec)
+
+    def _run_smoke_tests(self, _lib_spec: config.LibrarySpec) -> None:
```
miyuki, line 530: The test only supports armv6m, so I think it does not make sense to run the tests for each library variant we build.

scripts/make.py
```diff
@@ -526,3 +526,27 @@ def compiler_str(bin_name: str) -> str:
...
+            target = "run"
+        else:
+            logging.warning(
+                "qemu-arm is not present in your system path. Hence smoke "
+                "tests will only be built, not executed."
+            )
+            target = "build"
+        commands = ["make", target, f"BIN_PATH={bin_path}"]
+        try:
+            self.runner.run(commands, testdir_path)
```
miyuki, line 545: If we run the compiled binary, we should also check the output.

PR #36, opened 2021-05-05
Fix C++ static initialization

miyuki:
This patch changes the newlib build so that startup code calls
__libc_init_array (a function which traverses the .init_array section
and calls static object constructors).


vhscampos (approved):
LGTM

PR #37, opened 2021-05-06
Add tests to the build process

vhscampos:
This patch adds a 'test' phase to the build system.

For now we have added one smoke test, which is the same as
baremetal-semihosting from the 'samples' directory.

If qemu-arm is present in PATH, tests are built and run. Otherwise,
they are only built.

Output checking is done using golden files. Each smoke test must have
one stdout and one stderr file that contain the expected outputs, which
are in turn checked against the test output.

This current test framework requires 'make' installed and in PATH.

### DIFF
--- a/scripts/execution.py
+++ b/scripts/execution.py
@@ -38,14 +38,15 @@ class Runner:
         """
         self.last_cwd = None
 
-    def run(self, args: Sequence[str], cwd: str = None,
-            env: Mapping[str, str] = None) -> None:
-        """Run a specified program with arguments, optionally change current
-           directory and change the environment. Note: env does not replace
-           parent environment, but amends it for the subprocess."""
+    def _configure_env(self,
+                       args: Sequence[str],
+                       cwd: str = None,
+                       env: Mapping[str, str] = None) -> Mapping[str, str]:
         if env is not None:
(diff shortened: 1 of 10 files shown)


scripts/execution.py
```diff
@@ -78,18 +77,25 @@ def run(self, args: Sequence[str], cwd: str = None,
                               ex.returncode)
             raise
 
+        if self.verbose:
+            sys.stdout.write(result.stdout)
```
miyuki, line 81: This will have a side-effect: stdout and stderr will now be printed separately from each making the logs harder to interpret. Pease write a separate method for running a process with stdout and stderr capture.

scripts/execution.py
```diff
@@ -78,18 +77,25 @@ def run(self, args: Sequence[str], cwd: str = None,
...
+        if capture_stderr is not None:
+            capture_stderr[:] = result.stderr.decode('utf-8').splitlines()
 
 def run(args: Sequence[str], cwd: str = None, env: Mapping[str, str] = None,
         verbose: bool = False) -> None:
     """A wrapper for the Runner class which invokes run once."""
     runner = Runner(verbose)
     runner.run(args, cwd, env)
 
-
```
miyuki, line 88: Please remove this whitespace change and another on in the end of the file.

PR #39, opened 2021-05-11
Fixing the overwrite issue of multiple library builds

simpal01:
Currently every library variant we build goes into a folder differentiated
by llvm target triple.But when building consecutive libraries that have the
same target triple but differ in other attributes (ex. soft vs hard fp),
the files will overwrite each other in the same directory.

This issue is fixed by
    * Changing the default installation path from INSTALL_DIR/lib/clang-runtimes/TARGET
      to INSTALL_DIR/lib/clang-runtimes/LIBRARY_SPEC_NAME .
    * Also need to mention the sysroot to this new path.
    * Since the default sysroot is changed, need to add the library and header path in the
      configuration files.

### DIFF
--- a/scripts/cfg_files.py
+++ b/scripts/cfg_files.py
@@ -21,10 +21,10 @@ import config
 import util
 
 
-def copy_base_ld_script(cfg: config.Config, target: str) -> None:
+def copy_base_ld_script(cfg: config.Config, lib_spec_name: str) -> None:
     """Copy the linker script to target-specific directory."""
     base_ld_src = os.path.join(cfg.source_dir, 'ldscript', 'base.ld')
-    base_ld_dest = os.path.join(cfg.target_llvm_rt_dir, target, 'base.ld')
+    base_ld_dest = os.path.join(cfg.target_llvm_rt_dir, lib_spec_name, 'base.ld')
     if cfg.verbose:
         logging.info('Copying %s to %s', base_ld_src, base_ld_dest)
     shutil.copy(base_ld_src, base_ld_dest)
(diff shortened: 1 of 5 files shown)


scripts/make.py
```diff
@@ -534,6 +562,17 @@ def compiler_str(bin_name: str) -> str:
         if cfg.is_cross_compiling:
             self._copy_runtime_to_native(lib_spec)
 
+        # For newlib, the --with-target-subdir configure option is passed by default
```
stuij, line 565: I think it would be better when building the newlib files to output them in the build directory. as it is the files in 'target' are still there. and in general that gives us a bit more control over what actually ends up in the install directory.

scripts/make.py
```diff
@@ -534,6 +562,17 @@ def compiler_str(bin_name: str) -> str:
...
+        # Due to that target libraries are normally built in a subdirectory whose
+        # name is the argument to --target (ie, here it is getting built in the path
+        # "INSTALL_DIR/lib/clang-runtimes/{lib_spec_name}/{target}".
+        # Copying these built libraries and headers into the correct path
+        # "INSTALL_DIR/lib/clang-runtimes/{lib_spec_name}"
+
+        self._copy_newlib_headers_and_libs(join(cfg.target_llvm_rt_dir, lib_spec.name, lib_spec.target,                                                   'lib'), join(cfg.target_llvm_rt_dir, lib_spec.name, 'lib'))
```
stuij, line 572: the 'lib' here is all the way out on the same line. I'm guessing that should probably be on the next line or on the same line without the spaces in between

scripts/cfg_files.py
```diff
@@ -43,19 +43,25 @@ def write_cfg_files(cfg: config.Config, lib_spec: config.LibrarySpec) -> None:
 
     # No semihosting and no linker script
     nosys_lines = base_cfg_lines + [
-        '$@/../lib/clang-runtimes/{}/lib/crt0.o'.format(target),
+        '$@/../lib/clang-runtimes/{}/lib/crt0.o'.format(lib_spec.name),
+        '-L$@/../lib/clang-runtimes/{}/lib'.format(lib_spec.name),
```
miyuki, line 47: This line and the next one are common between all configs, so they should be moved to `base_cfg_lines`

scripts/cfg_files.py
```diff
@@ -43,19 +43,25 @@ def write_cfg_files(cfg: config.Config, lib_spec: config.LibrarySpec) -> None:
...
+        '-isystem $@/../lib/clang-runtimes/{}/include'.format(lib_spec.name),
```
miyuki, line 48: These two lines are common between all configs, so they should be moved to `base_cfg_lines`.

stuij:
this change has been superseded by: https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/pull/47

PR #46, opened 2021-05-14
Add automated checks for build scripts

miyuki:
This change:
- Adds a shell script to run precommit checks on Python scripts and
  mentions it in the README.md file
- Adds a Github workflow which runs the same checks on pull requests
  and updates on the 'main' branch

### DIFF
--- a/run-precommit-checks.sh
+++ b/run-precommit-checks.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+if [ ! -d ./venv ]; then
+    echo "Please run ./setup.sh"
+    exit 1
+fi
+
+. ./venv/bin/activate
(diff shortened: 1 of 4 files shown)


run-precommit-checks.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
vhscampos, line 1: Should we have this script portable instead?

vhscampos: OS-wise

miyuki: We are already using bash in setup.sh, so I guess it should be OK here as well.

vhscampos (approved):
LGTM

PR #47, opened 2021-05-18
Fixing the overwrite issue of multiple library builds

danhag01:
Currently every library variant we build goes into a folder differentiated
by llvm target triple.But when building consecutive libraries that have the
same target triple but differ in other attributes (ex. soft vs hard fp),
the files will overwrite each other in the same directory.

This issue is fixed by
    * Changing the default installation path from INSTALL_DIR/lib/clang-runtimes/TARGET
      to INSTALL_DIR/lib/clang-runtimes/LIBRARY_SPEC_NAME .
    * Also need to mention the sysroot to this new path.
    * Since the default sysroot is changed, need to add the library and header path in the
      configuration files.

Change-Id: Ie86564d00695f3c391953cf12f42b64cf829e5ed

### DIFF
--- a/samples/src/baremetal-semihosting/Makefile
+++ b/samples/src/baremetal-semihosting/Makefile
@@ -20,7 +20,7 @@ include ../../Makefile.conf
 build: hello.elf
 
 hello.elf: *.c
-	$(BIN_PATH)/clang --config armv6m-none-eabi_rdimon_baremetal -g -T ../../ldscripts/microbit.ld -o hello.elf ../../startup/startup_ARMCM0.S $^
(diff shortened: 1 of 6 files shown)


stuij (approved):
LGTM

stuij:
This is a continuation of: https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/pull/39

stuij:
One thing: please remove the `Change-id` line in the commit message before we merge this.

danhag01:
> One thing: please remove the `Change-id` line in the commit message before we merge this.

Should be fixed now.

PR #56, opened 2021-05-28
Fix check.py for Debian Clang

chommik:
Debian version of Clang is named 'Debian clang' in version string,
which fails the assertion that version string must start
with 'clang version'.

Also, relying on Clang version being the third word is no longer valid,
as there might be a 'Debian' prefix.

------------------

As for the backward compatibility:

```python3
In [1]: import re

In [2]: vers = [ 'clang version 10.0.0-4ubuntu1',
   ...: 'clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)',
   ...: 'clang version 3.8.0-2ubuntu4 (tags/RELEASE_380/final)',
   ...: 'Debian clang version 11.0.1-2', ]

In [3]: [ re.search(r'version (\S+?)(-.*)', v).group(1) for v in vers ]
Out[3]: ['10.0.0', '6.0.0', '3.8.0', '11.0.1']
```

### DIFF
--- a/scripts/check.py
+++ b/scripts/check.py
@@ -94,11 +94,14 @@ def _parse_clang_version(c_compiler: str) -> Optional[Version]:
     # Ubuntu 20.04: clang version 10.0.0-4ubuntu1
     # Ubuntu 18.04: clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
     # Ubuntu 16.04: clang version 3.8.0-2ubuntu4 (tags/RELEASE_380/final)
+    # Debian testing (bullseye): Debian clang version 11.0.1-2
     # Built from source: clang version 9.0.1
-    assert ver_line.startswith('clang version ')
-    ver_str = ver_line.split(' ')[2]
     # Remove distribution suffix (if any) and convert to a tuple
-    return _str_to_ver(ver_str.split('-')[0])
+    assert ver_line.startswith('clang version ') \
+        or ver_line.startswith('Debian clang version')
+    ver_match = re.search(r'version ([0-9.]+)', ver_line)
+    assert ver_match is not None
+    return _str_to_ver(ver_match.group(1))
 
 
 def _parse_gcc_version(c_compiler: str) -> Optional[Version]:


scripts/check.py
```diff
@@ -94,11 +94,14 @@ def _parse_clang_version(c_compiler: str) -> Optional[Version]:
     # Ubuntu 20.04: clang version 10.0.0-4ubuntu1
     # Ubuntu 18.04: clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)
     # Ubuntu 16.04: clang version 3.8.0-2ubuntu4 (tags/RELEASE_380/final)
+    # Debian testing (bullseye): Debian clang version 11.0.1-2
     # Built from source: clang version 9.0.1
-    assert ver_line.startswith('clang version ')
-    ver_str = ver_line.split(' ')[2]
     # Remove distribution suffix (if any) and convert to a tuple
-    return _str_to_ver(ver_str.split('-')[0])
+    assert ver_line.startswith('clang version ') \
```
miyuki, line 100: Please don't hardcode "Debian" here. Something like `'clang version ' in ver_line` should be fine.

scripts/check.py
```diff
@@ -94,11 +94,14 @@ def _parse_clang_version(c_compiler: str) -> Optional[Version]:
...
+        or ver_line.startswith('Debian clang version')
+    ver_match = re.search(r'version (\S+?)(-.*)', ver_line)
```
miyuki, line 102: This will not match "clang version 9.0.1" (from the "# Built from source:" comment). How about `r'version ([0-9.]+)'`?

miyuki (changes requested):

scripts/check.py
chommik: A few hours after I created this PR, out of curiosity, I've also tried to compile the LLVM-Embedded on macOS Big Sur:
```
% clang --version        
Apple clang version 12.0.5 (clang-1205.0.22.9)
Target: arm64-apple-darwin20.5.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```
and it fails with some nonsense errors:
```
clang-12: warning: argument unused during compilation: '-arch x86_64' [-Wunused-command-line-argument]
/Users/rafal/dev/LLVM-embedded-toolchain-for-Arm/repos-0.1/llvm.git/compiler-rt/lib/builtins/int_util.c:57:10: fatal error: 'stdlib.h' file not found
#include <stdlib.h>
         ^~~~~~~~~~
1 error generated.
```

I didn't have time to thoroughly look at the logs, so I'm not reporting an issue yet. In fact, the LLVM-Embedded repository doesn't mention `x86_64` in a meaningful way, so I'm not sure if that's even the correct repository to file a bug against.

Keep in mind, that Apple has a history of having their own LLVM versioning scheme, as per [wikipedia](https://en.wikipedia.org/wiki/Xcode#Xcode_7.0_-_12.x_(since_Free_On-Device_Development))

Therefore, of LLVM flavours, only "clang" and "Debian clang" are **definitely** compatible.

If not a hardcoded `Debian clang`, maybe a set of compatible clang flavours, like this?
```python
compatible_flavours = { '', 'Debian' }
```

scripts/check.py
chommik: Yes, that looks much cleaner. I forgot about the source build.
For a quick check:
```python
In [1]: import re

In [2]: vers = [ 'clang version 10.0.0-4ubuntu1',
   ...: 'clang version 6.0.0-1ubuntu2 (tags/RELEASE_600/final)',
   ...: 'clang version 3.8.0-2ubuntu4 (tags/RELEASE_380/final)',
   ...: 'Debian clang version 11.0.1-2',
   ...: 'clang version 9.0.1' ]

In [3]: [ re.search(r'version ([0-9.]+)', v).group(1) for v in vers ]
Out[3]: ['10.0.0', '6.0.0', '3.8.0', '11.0.1', '9.0.1']
```

scripts/check.py
miyuki: OK, good point about different versioning in Apple Clang. Let's leave the assertion as-is then. As for Mac support in general, we are currently focused on Linux and Windows and are not yet claiming that Mac is a supported host.

miyuki:
Please squash the two commits into a single commit, and I'll approve it.

chommik:
Done. Thanks for the review!

miyuki (approved):

PR #58, opened 2021-06-23
Add aarch64-none-elf 0.1 support

kbeyls:
This series of patches adds support for the aarch64-none-elf triple.
These patches introduce this support for the 0.1 version; but not yet for the HEAD version.
I thought it's going to be easier to first get the 0.1 version reviewed; the changes for the HEAD version should then be very similar and easy to review.

As this support should probably be considered experimental at the moment, I'm not sure if any documentation should be updated at all?


scripts/make.py
```diff
@@ -160,7 +160,7 @@ def build_native_tools(self) -> None:
         dist_comps += self.llvm_binutils
 
         cmake_defs = {
-            'LLVM_TARGETS_TO_BUILD:STRING': 'ARM',
+            'LLVM_TARGETS_TO_BUILD:STRING': 'ARM;AArch64',
```
miyuki, line 163: Please enable AArch64 conditionally (depending on `cfg.variants`).

scripts/make.py
```diff
@@ -160,7 +160,7 @@ def build_native_tools(self) -> None:
...
             'LLVM_DEFAULT_TARGET_TRIPLE:STRING': cfg.default_target,
```
miyuki, line 164: The `cfg.default_target is None` case needs to be handled here too (like in `build_clang`). Also, please consider factoring out common logic related to `cmake_defs` from `build_native_tools` and `build_clang`.

miyuki (commented):
A few questions and requests:
1. Does the build succeed after each commit? I.e. shouldn't the "Fix cmake configuration issues" and "set CMAKE_SYSTEM_NAME to 'Generic'" go before "Add (experimental) aarch64-none-elf support"?
2. Does this patch series break the HEAD build? If it does please disable AArch64 when building the HEAD version.
3. Please add armv8a-64 to the "Supported architectures" sections in README.md and mention that the support is experimental.

scripts/make.py
kbeyls: Will do. I assume that both the ARM and AArch64 backends should be made conditional, based on whether they are needed.

scripts/make.py
kbeyls: Thanks, good catch.

kbeyls:
> 
> 
> A few questions and requests:
> 
>     1. Does the build succeed after each commit? I.e. shouldn't the "Fix cmake configuration issues" and "set CMAKE_SYSTEM_NAME to 'Generic'" go before "Add (experimental) aarch64-none-elf support"?

I wasn't sure if keeping each build succeeding was important in this project or not.
I'll follow your guidance here and verify builds keep on succeeding after each commit for the next version of this patch series.
 
>     2. Does this patch series break the HEAD build? If it does please disable AArch64 when building the HEAD version.

It does. Maybe the best way to fix this is for every commit that changes the .patch files (which are version-specific), to change both the the -0.1 and -HEAD versions in the same commit?

>     3. Please add armv8a-64 to the "Supported architectures" sections in README.md and mention that the support is experimental.

Will do, thanks for the guidance.

scripts/make.py
miyuki: IIRC, an AArch64-only configuration will fail at build time because the AArch64 backend uses some functions from the Arm backend. However, I am not 100% certain.

miyuki:
> I wasn't sure if keeping each build succeeding was important in this project or not.

We already had at least one bug where `git bisect` turned out to be very helpful, so I think it is important.

> It does. Maybe the best way to fix this is for every commit that changes the .patch files (which are version-specific), to change both the the -0.1 and -HEAD versions in the same commit?

That would be ideal, but if you don't want to spend time on creating patches for the HEAD version right now, disabling AArch64 for HEAD is also fine.

scripts/make.py
kbeyls: I implemented both ARM and AArch64 backends to be conditional. I didn't see any failures when the ARM backend wasn't present for an aarch64-only toolchain.

scripts/make.py
kbeyls: This has been fixed/implemented (including the factoring out common logic) in the updated patch series.

kbeyls:
I've now updated the patch series, addressing all previous comments.

> > I wasn't sure if keeping each build succeeding was important in this project or not.
> 
> We already had at least one bug where `git bisect` turned out to be very helpful, so I think it is important.

I tested that both the 0.1 and HEAD version builds work correctly after each commit.

> 
> > It does. Maybe the best way to fix this is for every commit that changes the .patch files (which are version-specific), to change both the the -0.1 and -HEAD versions in the same commit?
> 
> That would be ideal, but if you don't want to spend time on creating patches for the HEAD version right now, disabling AArch64 for HEAD is also fine.

kbeyls:
Oh, and I also added aarch64, armv8.0 to the "Supported architectures" sections in README.md and mentioned that the support is experimental.

miyuki (approved):
LGTM

PR #59, opened 2021-07-12
aarch64-none-elf: add ldscript and config file.

kbeyls:
This adds generation of .cfg files for the aarch64-none-elf
triple; including a base linker script.


miyuki (approved):
LGTM

PR #60, opened 2021-08-05
Update the LLVM patch (HEAD version)

miyuki:
The current version no longer applies cleanly because of the changes
in the libc++ math.h header related to the implementation of the
copysign functions.


voltur01:
Thanks for looking into this! I can confirm that the old patch did not apply, while the updated patch applies cleanly and the toolchain builds successfully afterwards.

vhscampos (approved):
LGTM

PR #62, opened 2021-09-19
Support build on macOS

mbrossard:
I can report a successful build with this small patch on macOS X with `clang` version from Xcode Command Line Tools (which includes `make` as well) and additional software components from brew.

```
$ brew install cmake ccache ninja python virtualenv gnu-tar
$ ./setup.sh
$ . ./venv/bin/activate
$ ln -s /usr/local/bin/gtar venv/bin/tar
$ build.py
```

Notes:
- `build.py` assumes `tar`command is GNU version, simplest work-around is to install via brew and use symbolic link.
- With the resulting binaries I was able to build the smoke tests but I did not run them.


mbrossard:
I created a small GtitHub Actions workflow to show that it works, you can see it in action [here](https://github.com/mbrossard/LLVM-embedded-toolchain-for-Arm/actions/workflows/build-macos.yml).

kbeyls:
I personally think it would be better to not rely on GNU tar features if possible rather than require GNU tar to be installed.
My guess is that for developers on non-GNU unix-like platforms that will be easier.
What is the GNU tar specific feature that is being relied on, and is it really necessary?

mbrossard:
With the native `tar` (`bsdtar 3.3.2`) I get the error:

```ERROR: Command failed with return code 1, stderr:
tar: Option --owner=root is not supported```

It seems they have incompatible names for the options to set owner / group (`--uname` and `--gname`).

miyuki:
I've raised #64 to track the issue with `bsdtar`.

PR #63, opened 2021-10-05
Use fixed versions of PyYAML and GitPython

miyuki:
This patch adds a new file requirements.txt that contains dependencies
of build scripts with fixed versions. The patch fixes Python 3.9 lint
failure.


amilendra (approved):
LGTM

PR #76, opened 2021-11-04
Build libunwind instead of a dummy one.

DanielKristofKiss:
libunwind is a perquisite to have exceptions for example.

### DIFF
--- a/scripts/make.py
+++ b/scripts/make.py
@@ -286,6 +286,7 @@ class ToolchainBuild:
             'CMAKE_C_COMPILER': join(cfg.native_llvm_bin_dir, 'clang'),
             'CMAKE_C_COMPILER_TARGET': lib_spec.target,
             'CMAKE_C_FLAGS': flags,
+            'CMAKE_ASM_FLAGS': flags,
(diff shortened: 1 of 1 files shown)


mplatings:
We're not planning to support features like exceptions so we've decided to leave libunwind disabled.

mplatings:
> We're not planning to support features like exceptions
This isn't quite right. We don't have any immediate plans to build library variants with exception support, but that doesn't mean this pull request shouldn't be accepted.

mplatings:
The updated change is failing to build:

> clang -cc1as: error: unknown target triple 'x86_64-unknown-linux-gnu', please use -triple or -arch

DanielKristofKiss:
@mplatings it built for me with todays ToT llvm.

mplatings:
I'm not sure why the build succeeded for you. I dug in deeper and it turned out that .S files weren't building because CMAKE_ASM_FLAGS wasn't specified. I pushed a fix for that.
The smoke tests are currently failing for me so I'm going to investigate that before returning to this.

PR #77, opened 2021-11-05
Create a shallow clone of the dependent repositories.

DanielKristofKiss:
No need for the whole history so let's clone only the requested
revision. This speeds up the clone significantly and reduce the required
disk space.

### DIFF
--- a/scripts/repos.py
+++ b/scripts/repos.py
@@ -263,23 +263,21 @@ def clone_repositories(checkout_path: str, tc_version: LLVMBMTC,
         logging.info(' - %s: %s @ %s%s', repo_path, module.branch,
                      module.revision,
                      ' (detached)' if module.revision != 'HEAD' else '')
+        refspec = (module.branch if module.revision == 'HEAD'
+                   else module.revision)
+
         repo = git.Repo.clone_from(module.url,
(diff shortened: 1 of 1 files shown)


scripts/repos.py
```diff
@@ -263,23 +263,21 @@ def clone_repositories(checkout_path: str, tc_version: LLVMBMTC,
...
-            # Detached state
-            try:
-                repo.git.checkout(module.revision)
-            except git.exc.GitCommandError as ex:  # pylint: disable=no-member
-                die('could not checkout "{}" @ "{}".\n'
-                    'Git command failed with:\n{}'
-                    .format(repo_path, module.revision, ex))
+                                   os.path.join(checkout_path, module.name),
+                                   multi_options=[
+                                       "--branch %s" % (refspec),
```
miyuki, line 272: Does this work with tags and commit hashes? E.g. does `build.py -r 13.0.0` work?

DanielKristofKiss: It works only with with tags and branches.
`build.py -r 13.0.0` works.

miyuki: OK. Since we are currently not using commit hashes in versions.yml, supporting just tags and branches should be fine.

miyuki (approved):

PR #82, opened 2021-12-02
Add Mingw-w64 licenses and version information

miyuki:
This patch adds several license files pertinent to the runtime DLLs
that are part of the Mingw-w64 project, specifically:
* COPYING3.GCC - GPL v3 - applies to libstdc++ and libgcc
* COPYING.RUNTIME - GCC Runtime library exception - applies to
  libstdc++ and libgcc
* COPYING.MinGW-w64-runtime.txt - applies to libstdc++, libgcc and
  libwinpthread

These licenses are included in the binary distribution when the
corresponding DLLs are included in the distribution.

Also this patch adds Mingw-w64 version information to VERSION.txt
when compiling with Mingw-w64.


scripts/package.py
```diff
@@ -42,6 +42,17 @@ def _write_version_file(cfg: config.Config, version: repos.LLVMBMTC,
     util.write_lines(lines, dest)
 
 
+def _maybe_append_mingw_version(cfg: config.Config, target_dir: str) -> None:
```
walkerkd, line 45: Would this be better named _append_mingw_version_if_included ?

walkerkd (changes requested):

walkerkd (approved):

PR #90, opened 2022-03-29
Fix HEAD build

miyuki:
These patches are the first in a series. The goal of the series is to fix the build failure of the HEAD version.


voltur01 (approved):
LGTM, I can build HEAD now and the tests pass.

PR #91, opened 2022-03-30
Use new syntax to specify bin directory in config files

miyuki:
Upstream patch https://reviews.llvm.org/D115604 implemented a feature
that we used to implement in a downstream patch. Config files can now
use a special placeholder that is automatically replaced with the
directory containing the config file. The placeholder used by upstream
is different from the downstream one ('<CFGDIR>' vs '$@'), this
requires some adjustments in the python script that generates the
config files.


voltur01 (approved):
Thank you - great improvement! The change LGTM, builds successfully and runs the tests.

PR #92, opened 2022-03-31
Fix HEAD build, two more patches

miyuki:
The patches update llvm-HEAD.patch to accommodate upstream changes in libc++ headers.


voltur01 (approved):
LGTM and applies cleanly.

PR #93, opened 2022-03-31
Use new method to build libc++abi and libc++

miyuki:
Upstream change https://reviews.llvm.org/D119255 drops support for
standalone builds of libc++abi and libc++. The new way to build these
libraries is to invoke CMake using the source root directory
'runtimes', not 'libcxx' or 'libcxxabi'.

This patch updates the scripts to build libc++ and libc++abi
correctly, the additional benefit is that libc++ and libc++abi are now
built in one go.

Unfortunately, compiler-rt cannot be built in the same way because it
is incompatible with the workaround that involves manually setting
CMAKE_SYSTEM_NAME (which is needed for AArch64, see commit
cd997482aff515872b9938a9d529ee2b2447ae29).

### DIFF
--- a/scripts/make.py
+++ b/scripts/make.py
@@ -281,10 +281,6 @@ class ToolchainBuild:
                                                lib_spec.name)))
         defs = {
             'CMAKE_TRY_COMPILE_TARGET_TYPE': 'STATIC_LIBRARY',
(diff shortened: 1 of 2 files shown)


voltur01 (approved):
LGTM

PR #95, opened 2022-04-04
Add build configurations for LLVM 14

miyuki:
This patch adds two build configurations for LLVM 14: one for tip of
LLVM 14 branch and one for LLVM 14.0.0 release. Both configurations
use Newlib 4.1.0 (the latest stable release).

The patch also removes LLVM 13 configurations because after commit
036932532c0e820a08f05d4f0447bd0311b546c9 the config files generated by
the build script are no longer compatible with LLVM 13. One should use
the llvm-13 branch to build a toolchain based on LLVM 13.


voltur01 (approved):
Looks good, builds and runs tests cleanly.

PR #98, opened 2022-04-05
Update newlib-HEAD.patch, remove libgloss/libnosys/configure

miyuki:
Upstream newlib commit d5a20f0b70c73c72ec2bc4b639815bb821859255
merges libgloss/libnosys/configure into libgloss/configure. The new
configure script uses a proper check for ELF support rather than
checking that the target triple ends with "-elf" (which gives
incorrect result for "arm-*-eabi").

Because of the change we no longer need a downstream patch for the
configure script.


voltur01 (approved):
Thank you for the update, looks good!

PR #99, opened 2022-04-07
Fix newlib breakage related to INIT_FINI

miyuki:
The newlib change 437c5c5085ff30b4a4960b2b53d06728c788361d renamed
HAVE_INIT_FINI to _HAVE_INIT_FINI and HAVE_INITFINI_ARRAY to
_HAVE_INITFINI_ARRAY.

This patch adjusts the build scripts by adding -U_HAVE_INIT_FINI to
compiler flags (alongside with -UHAVE_INIT_FINI which we need to keep
in order to build older newlib versions).

The patch also renames HAVE_INITFINI_ARRAY to _HAVE_INITFINI_ARRAY in
assembly files (this part should be upstreamed at some point).


voltur01 (approved):
This solves the init problem, thank you!

PR #100, opened 2022-04-07
Update newlib-HEAD.patch, startup code

miyuki:
Upstream newlib commit 063d67faf0266e5ba169bc5cfde8aed011b1d41b fixes
the issue with HAVE_INITFINI_ARRAY (the macro had been renamed in
configure scripts and C source code, but not in startup code written
in assembly), so we no longer need a downstream fix for it.


voltur01 (approved):
LGTM

PR #101, opened 2022-04-08
Remove hard-coded commit hashes from versions.yml HEAD config

miyuki:
A prior commit b4aa4b716dc3c1b4923b57c048b5e0e79c13ca5c changed the
HEAD build configuration in versions.yml to use hard-coded commits of
LLVM and newlib, so that build breakages could be fixed one by one.

Now that all breaking changes have been addressed, we can restore the
HEAD configuration to track the tips of upstream branches as before.


voltur01 (approved):
LGTM

PR #114, opened 2022-07-28
Build newlib with -Wno-implicit-function-declaration

mplatings:
Newlib only declares long double functions like cosl if long double is
equivalent to double. On AArch64 long double is twice the size of double
so the declarations are missing. However Newlib's build system still
tries to compile the long double code.

The LLVM change https://reviews.llvm.org/D122983 means that clang no
longer implicitly declares functions by default. Therefore use
-Wno-implicit-function-declaration to work around the problem.

Ideally Newlib's build system would be fixed, but since we plan on
switching to picolibc (which doesn't have this problem) just use this
workaround instead of fixing it properly.


miyuki (approved):
LGTM

PR #116, opened 2022-08-01
Remove support for building version 14

mplatings:
If you wish to build version 14 then please use the llvm-14 branch.

The --revision option is retained as it is likely to still be
occasionally useful. However in general using the relevant branch of
the repo is preferred so the option now has less prominence in the
documentation.


miyuki (approved):
LGTM

Issue #61, opened 2021-09-18
picolibc instead of newlib


beriberikix:
Very excited to see this happening! My first question is have you evaluated [picolibc](https://github.com/picolibc/picolibc) instead of newlib? Differences between newlib(-nano) are discussed  in picolibc/picolibc#103.

miyuki:
No, we haven't, but we will have a look into it.

mplatings:
https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/pull/120

mplatings:
Re-landed in #123 
Smoke tests are passing but samples still need updating

mplatings:
Samples have been updated. I believe this is done.

PR #117, opened 2022-08-01
Replace newlib with picolibc

mplatings:
GitHub issue #61


versions.yml
```diff
@@ -26,9 +26,9 @@ Revisions:
         Branch: main
         Revision: HEAD
         Patch: llvm-HEAD.patch
-      - Name: newlib.git
+      - Name: picolibc.git
         FriendlyName: Newlib
```
miyuki, line 30: It guess this should say Picolibc

scripts/package.py
```diff
@@ -153,10 +153,10 @@ def add_multiple_licenses(comp_name, paths):
...
         add_license(comp_name, src_path, dest_name)
 
-    newlib_lic_names = ['COPYING.NEWLIB', 'COPYING.LIBGLOSS']
-    newlib_lic_files = [os.path.join(cfg.newlib_repo_dir, n)
-                        for n in newlib_lic_names]
-    add_multiple_licenses('Newlib', newlib_lic_files)
+    picolibc_lic_names = ['COPYING.NEWLIB', 'COPYING.picolibc']
+    picolibc_lic_files = [os.path.join(cfg.picolibc_repo_dir, n)
+                          for n in picolibc_lic_names]
+    add_multiple_licenses('Newlib', picolibc_lic_files)
```
miyuki, line 159: Should be 'Picolibc' instead of 'Newlib'

versions.yml
miyuki: FriendlyName still says Newlib

miyuki: Oh, I see it's in a different commit.

docs/building-from-source.md
```diff
@@ -146,9 +146,7 @@ $ build.py --host-toolchain mingw --copy-runtime-dlls no
 
 ## Divergences from upstream
 
-### newlib:
-* Clang does not support the ``naked`` attribute on C functions, breaking the
-  Linux startup (out of scope).
+### picolibc:
 * Target triple ending with eabi is not considered an ELF target.
```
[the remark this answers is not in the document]
mplatings: I have no idea.

I got rid of the "breaking Linux startup" bullet because it looks like picolibc doesn't have such startup. But I don't understand the "Target triple ending with eabi is not considered an ELF target" bullet so I can't judge whether it still applies.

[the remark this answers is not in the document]
mplatings: Removed. Now only the patch to enable building with -mthumb is described here.

miyuki (approved):

PR #118, opened 2022-08-02
Allow building with CMake release candidate

mplatings:


scripts/check.py
```diff
@@ -58,6 +58,10 @@ def __lt__(self, other):
 
 def _str_to_ver(version_str: str) -> Version:
     """Covert version string to a tuple of numbers."""
+    # Strip pre-release and build metadata
```
miyuki, line 61: We already have some program-specific preprocessing in `_check_tool`, `_parse_clang_version` and `_parse_gcc_version`, so I think this code belongs to `_check_tool`.

mplatings: This isn't program-specific. "1.2.3-rc4+build5" is a valid version string according to semantic versioning therefore the function should handle it, no?

miyuki (approved):

PR #120, opened 2022-08-05
Revert "Replace newlib with picolibc"

mplatings:
The switch caused a regression in semihosting support which will take
time to fix. The patch will be relanded once the fix is in place.

This reverts commit 1e08e801d5c5020e6a2b8b17b070e53379d3250d.


amilendra:
LGTM.

PR #121, opened 2022-08-05
Move default install location into build dir

mplatings:
The smoke tests use the install dir, which meant that deleting the build
dir didn't assure a valid test result.
When compared to other build systems this is surprising.
By moving the install dir into the build dir the expectation that
deleting the build dir will cause tests to use fresh build output is
upheld.

For consistency, also default to putting the package tar.gz in the build
directory.


statham-arm (approved):
LGTM.

Certainly the package .tar.gz is a build product, so it makes sense to put it inside the build dir with the rest of the build products.

Putting the 'install' directory in there sounds slightly more surprising to people who were expecting `make install` (or equivalent) to actually _install_ things (e.g. default to `/usr/local`), but the previous version didn't do that either, so it's clear that in this context the "install" directory just means "somewhere you can run the tools from right now for evaluation purposes, no implication of installing systemwide".

PR #123, opened 2022-08-10
Picolibc

mplatings:
GitHub issue #61

* Restore picolibc by reverting #120
* Rename rdimon to semihost
* Turn off wide character support in libc++ to make it play nicely with picolibc's tinystdio.
* Add a mbstate_t header to picolibc in the place that libc++ expects to find it when wide character support is disabled.
* The ldscript/base_arm.ld linker script doesn't work with picolibc. Maybe it should be removed entirely but that's a discussion for another time. For now add the minimal arm.ld in the smoketests directory.
* Picolibc's semihosting uses a single FILE object for stdout/err so we can't discriminate between the two.

I intend to squash these changes once approved.


statham-arm (approved):
By diffing against the state just before the previous revert, I managed to turn this into a manageably small review :-) The changes from the previous picolibc patch LGTM.

Issue #109, opened 2022-04-15
--print-libgcc-file-name points to a wrong location


pkarashchenko:
```
clang  --config armv8m.main_hard_fp_nosys --target=armv8m.main-none-eabi -mfloat-abi=hard  -march=armv8m.main+fp --print-libgcc-file-name

/opt/install/LLVMEmbeddedToolchainForArm-13.0.0/lib/clang/13.0.0/lib/baremetal/libclang_rt.builtins-armv8m.main.a
```
but file does not exist.

voltur01:
Closing as per the above commit.

PR #128, opened 2022-08-24
Introduce -resource-dir flag to config files

mblenczewski-arm:
Running `clang --config <arch>.cfg --print-libgcc-file-name` now returns
a correct path for a compiler-rt builtins archive for the given
architecture.

GitHub issue #109

### DIFF
--- a/scripts/make.py
+++ b/scripts/make.py
@@ -296,6 +296,7 @@ class ToolchainBuild:
         }
         return defs
 
+    # pylint: disable=too-many-locals
     def build_compiler_rt(self, lib_spec: config.LibrarySpec) -> None:
         """Build and install a single variant of compiler-rt."""
         self.runner.reset_cwd()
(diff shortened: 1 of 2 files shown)


scripts/make.py
```diff
@@ -542,7 +536,6 @@ def build_picolibc(self, lib_spec: config.LibrarySpec) -> None:
             crossfile_path,
             cfg.picolibc_repo_dir,
         ]
-
```
mplatings, line 545: put this back

mplatings:
Change looks good in theory but can you add a comment in the code to make it clearer why the magic "baremetal" directory is the right one?
The linter failed but I'm OK with adding a special comment to persuade it that your variable is OK https://stackoverflow.com/a/23542817/2651243

mplatings:
See the contributing guide for how to make sure the linter and other checks pass: https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/blob/main/docs/contributing.md

mplatings (approved):
LGTM

PR #131, opened 2022-09-13
Remove incorrect modifications to picolibc.ld

mplatings:
The modifications happened to make tests pass, but broke binaries running on real hardware.

The tests were using qemu-arm, a QEMU User space emulator. Having binaries work on qemu-arm is a nice-to-have, but not a priority for the project. Therefore this change switches to qemu-system-arm, which emulates real hardware.

This change breaks the user space semihosting sample. Since supporting qemu-arm is not a priority the sample is removed, although it may be restored in future.


voltur01 (approved):
LGTM, thank you!

PR #132, opened 2022-09-13
Stabilise baremetal-uart sample

mplatings:
When run on micro:bit v1 hardware it lost most of the characters - now fixed.


voltur01 (approved):
LGTM

PR #135, opened 2022-09-23
.cfg suffix is no longer automatically added to --config

mtomczykmobica:
Since https://reviews.llvm.org/D134208 --config armv6m_soft_nofp_semihost no longer works because the feature to automatically append .cfg was removed. This breaks the smoke tests and the documented behaviour of BMT. Fix by added .cfg suffix to every --config option in documentation and examples.


mplatings:
LGTM. One more thing - please document the change in CHANGELOG.md

mtomczykmobica:
@mplatings Done, I've modified also CHANGELOG.md.

mtomczykmobica:
Locally tested - smoke test passed.

PR #136, opened 2022-09-27
 Fix "undefined symbol: operator new"

mplatings:
Clang's default C++ standard is now gnu++17 instead of gnu++14.
Supporting C++17 requires aligned new & delete. Implementing this
requires posix_memalign. Picolibc only provides posix_memalign if the
correct feature test macros are set.
https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -418,7 +418,10 @@ while(library_variants)
 
     # compiler-rt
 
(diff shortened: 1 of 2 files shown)


voltur01 (approved):
LGTM

PR #140, opened 2022-10-11
Adjust revisions and patches for LLVM 15

mplatings:


versions.yml
```diff
@@ -18,17 +18,17 @@
 ---
 SourceType: "standalone-build-scripts"
 Revisions:
-  - Revision: HEAD
```
miyuki, line 21: Could you please instead create a branch and update versions.yml on the branch.

voltur01 (approved):
LGTM

versions.yml
mplatings: Using llvm-15 branch

PR #147, opened 2022-10-20
Add infrastructure for using different C libraries.

statham-arm:
This commit adds a variable 'libc_type' set at the start of the loop over library variants, and moves the picolibc build subproject definition into an if statement that conditions on libc_type. So it's easy to apply a downstream patch that varies libc_type (perhaps per library variant) and adds an extra clause to provide a build subproject for the alternative library.

I've moved a lot of the LLVM runtimes configuration options into a variable 'libc_specific_runtimes_options', which is set inside the if statement that tests libc_type. Rationale: a different libc might provide a different set of functionality, and permit a different set of libc++'s optional pieces to be enabled.

No functional change. As long as libc_type is unconditionally set to 'picolibc', which it is in this commit, everything should build exactly as before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -401,31 +401,50 @@ while(library_variants)
         set(MESON_INSTALL_QUIET "--quiet")
     endif()
 
-    # picolibc
-    ExternalProject_Add(
-        picolibc_${variant}
-        SOURCE_DIR ${picolibc_SOURCE_DIR}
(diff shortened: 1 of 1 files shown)


mplatings:
The file currently uses indentation of 4 spaces so let's stay consistent within the file. I'd be supportive of a separate commit to reindent it with 2 spaces for consistency with llvm-project.

mplatings:
LGTM

PR #148, opened 2022-10-25
Remove Python build scripts

mplatings:


.cmakelintrc
```diff
@@ -0,0 +1 @@
+filter=-linelength,-readability/wonkycase
```
rockdreamer, line 1: I would highly recommend the use of consistent command case

mplatings: Unfortunately it objects to things like FetchContent_Declare which is taken directly from the official [docs](https://cmake.org/cmake/help/latest/module/FetchContent.html).
Maybe this could be changed in future but for now I'd prefer to have the settings match the style of the code as it is.

statham-arm (approved):

PR #154, opened 2022-11-11
Added a note on sandboxing the toolchain

voltur01:


mplatings:
Maybe put this in the docs folder? It seems a bit niche to feature on the front page of the project

voltur01:
Now this is the only place in our docs where we refer to usage of the toolchain, this was my logic. 

I can put a link in this usage section to a separate page in the docs folder, however just one paragraph for now seems a bit short for a whole new page - we may extend it with other security related information in the future, I guess.

voltur01:
Alternatively, I can shorten it to one sentence not to overload the main page:

> *Note:* If you are using the toolchain in a shared environment with untrusted input,
> make sure it is sufficiently sandboxed.

mplatings:
The shortened note SGTM. I think it's actually better because it doesn't imply "do these things and you'll be safe" but instead puts the responsibility on the reader to figure out what "sufficiently sandboxed" means for them.

voltur01:
Thank you, wording updated.

PR #155, opened 2022-11-14
Update C++ support info in README

mplatings:
Also update the change log. Support for C++17's aligned operator new was added with the switch to picolibc.


voltur01 (approved):
LGTM, thank you!

PR #156, opened 2022-11-17
Add experimental support for OSX host

stuij:
Just a tiny tweak was needed to get building the toolchain for OSX work: For the runtimes component, which builds libcXX*, including libunwind, we need to disable the APPLE CMake variable that will configure various bits for an assumed Apple target. It doesn't seem possible to disable this without directly editing the source.

Testing has been restricted to building the variants, running the `ninja check-llvm-toolchain' script (qemu runs produce the expected results) and doing some ad-hoc compiling for the targets.

Preferably we should handle this properly in LLVM proper, but accounting for the APPLE variable in a non-hacky way seems like a hairy affair, which will take more effort. Patching the source here seems like an acceptable band-aid for supporting a platform that we have had multiple requests for.

For a bit more detail, see the added comment in the CMakeLists.txt diff.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,9 @@ see component links above.
 The LLVM Embedded Toolchain for Arm has been built and tested on Linux/Ubuntu
 18.04.5 LTS, including cross-building for Windows.
 
+Building on macOS is functional but experimental. Currently it is only lightly tested on
+a Macbook Pro with M1 on macOS 12.3.1. Cross-building for Windows from Mac is untested.
+
 [Binary packages](https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/releases) 
 are provided for major LLVM releases for Linux and Windows.
 
--- a/docs/building-from-source.md
+++ b/docs/building-from-source.md
@@ -5,25 +5,31 @@
 The LLVM Embedded Toolchain for Arm has been built and tested on Linux/Ubuntu
 18.04.5 LTS.
 
+Building and testing on macOS is functional but experimental.
+
 ## Installing prerequisites
 
 The build requires the following software to be installed, in addition
-to the [LLVM requirements|https://llvm.org/docs/GettingStarted.html#software]:
+to the [LLVM requirements](https://llvm.org/docs/GettingStarted.html#software):
 * CMake 3.20 or above
 * Meson
 * Git
 * Ninja
 * Python
+* QEMU (for running the test suite, so optional)
 
 On a Ubuntu 18.04.5 LTS machine you can use the following commands to install
 the software mentioned above:
 ```
+$ apt-get install python3 git make ninja-build qemu
 $ apt-get install clang # If the Clang version installed by the package manager is older than 6.0.0, download a recent version from https://releases.llvm.org or build from source
-$ apt-get install python3
-$ apt-get install git
-$ apt-get install make
-$ apt-get install ninja-build
 $ apt-get install cmake # If the CMake version installed by the package manager is too old, download a recent version from https://cmake.org/download and add it to PATH
+$ pip install meson
+```
+
+On macOS, you can use homebrew:
+```
+$ brew install llvm python3 git make ninja qemu cmake
 $ pip install meson
 ```
 


README.md
```diff
@@ -66,6 +66,9 @@ see component links above.
 The LLVM Embedded Toolchain for Arm has been built and tested on Linux/Ubuntu
 18.04.5 LTS, including cross-building for Windows.
 
+Building on OSX is functional but experimental. Currently it is only lightly tested on
```
mplatings, line 69: Nit: OSX -> macOS

README.md
```diff
@@ -66,6 +66,9 @@ see component links above.
...
+a Macbook pro M1 with OSX 12.31. Cross-building for Windows is untested.
```
mplatings, line 70: Rephrase slightly to "Cross-building for Windows from Mac is untested"

patches/llvm-HEAD.patch
```diff
@@ -618,3 +618,35 @@ index ac57360cca9e..c8edc8bc8f21 100644
...
++
+ project(Runtimes C CXX ASM)
+ 
++# The CMake documentation on the APPLE variable reads:
++#   Set to True when the target system is an Apple platform (macOS, iOS, tvOS or
++#   watchOS).
++#
++# So this variable is used as a Target, however on Apple host platforms it is
++# automatically set to true, and it seems you can't change it through say the
++# command line your generator doesn't support platform specification, which
```
mplatings, line 637: This surprises me. What happens if in LLVM-embedded-toolchain-for-Arm/CMakeLists.txt you add -DAPPLE=False to the CMAKE_ARGS for runtimes?

README.md
voltur01, line 70: Style: I think their official spelling would be "MacBook Pro with M1"

patches/llvm-HEAD.patch
stuij: Yes I tried that. It doesn't register. It also doesn't register if in the runtimes CMakeLists.txt file I set APPLE to False before `project(Runtimes C CXX ASM)`, so I'm assuming that it's set somewhere in there.

I'm by far a CMake expert, and after I googled around for far too long, the best I got was setting the CMAKE_GENERATOR_PLATFORM variable. But as I noted in the code comment, this doesn't work for ninja as it doesn't support setting the platform specification.

It does baffle me and I'm assuming/hoping I missed something obvious, because why wouldn't you be able to change a variable that according to the documentation indicates which platform you're building for.

stuij: ping

mplatings: I guess that makes sense since it's a variable that CMake defines. However modifying CMake internal state might be asking for trouble. Since this change isn't urgent I'd prefer we take longer and find a non-hacky fix.

voltur01: I agree, this actually looks like a candidate for https://cmake.org/cmake/help/book/mastering-cmake/chapter/Cross%20Compiling%20With%20CMake.html - do we need to specify the toolchain and maybe even reuse it (BMT toolchain definition) on other platforms for building libraries?

stuij: I actually read that page, and tried to fix the issue with suggestions therein, but nothing helped. You can set `CMAKE_SYSTEM_NAME` to Generic, but then CMAKE will fail checks at least for embedded targets as it does certain LLVM checks that need host system support. And it might still set APPLE to True. Didn't check that.

I'm not arguing that we shouldn't solve this in a non-hackish way, but my guess is that this needs a saner setup on the LLVM side, which would need a bit more effort.

stuij:
updated to work with HEAD, added QEMU dependency

patches/llvm-HEAD.patch
mplatings: > You can set `CMAKE_SYSTEM_NAME` to Generic, but then CMAKE will fail checks at least for embedded targets

Should be fixed by https://reviews.llvm.org/D144757. Review please!

mplatings:
I landed #188 which I think allows building on Mac (specifically the addition of `-DCMAKE_SYSTEM_NAME=Generic`), please let me know if it works for you

stuij:
Yay, that's great!

I've tested this locally. Building and testing works with latest llvm and Embedded Toolchain.

I've updated this pull request to just add OSX host support information to the documentation.

docs/building-from-source.md
```diff
@@ -5,28 +5,34 @@
 The LLVM Embedded Toolchain for Arm has been built and tested on Linux/Ubuntu
 18.04.5 LTS.
 
+Building and testing on OSX is functional but experimental.
```
mplatings, line 8:
```suggestion
Building and testing on macOS is functional but experimental.
```

docs/building-from-source.md
```diff
@@ -5,28 +5,34 @@
...
 $ apt-get install clang # If the Clang version installed by the package manager is older than 6.0.0, download a recent version from https://releases.llvm.org or build from source
-$ apt-get install python3
-$ apt-get install git
-$ apt-get install make
-$ apt-get install ninja-build
 $ apt-get install cmake # If the CMake version installed by the package manager is too old, download a recent version from https://cmake.org/download and add it to PATH
 $ pip install meson
 ```
 
+On OSX, you can use homebrew:
```
mplatings, line 30:
```suggestion
On macOS, you can use homebrew:
```

README.md
```diff
@@ -69,6 +69,9 @@ see component links above.
 The LLVM Embedded Toolchain for Arm has been built and tested on Linux/Ubuntu
 18.04.5 LTS, including cross-building for Windows.
 
+Building on OSX is functional but experimental. Currently it is only lightly tested on
```
mplatings, line 72:
```suggestion
Building on macOS is functional but experimental. Currently it is only lightly tested on
```

mplatings (approved):

PR #161, opened 2022-11-30
Remove LLVM patch

mplatings:
Since https://reviews.llvm.org/D136393 we no longer need to patch LLVM.

Rather than removing the patching machinery, replace the patch with a dummy just in case we need to patch again in future.


statham-arm (approved):
I know I normally complain about diffs-to-diffs being amazingly hard to review, but even I'm prepared to admit that this one is easy to understand :-) LGTM!

PR #163, opened 2022-12-08
Group library variants by aarch64-none-elf & arm-none-eabi

mplatings:
Grouping them in this way is an enabler for implementing multilib.


CMakeLists.txt
```diff
@@ -357,9 +339,58 @@ foreach(variant ${LLVM_TOOLCHAIN_LIBRARY_VARIANTS})
...
 
-# Destructively iterate over the 3-tuples defining the variants.
-while(library_variants)
-    list(POP_FRONT library_variants target_arch variant_suffix flags)
+# Define which library variants to build and which flags to use.
+# The order is <parent directory name> <libc_type> <arch> <variant name> <flags>
+# e.g. "arm-none-eabi" "picolibc" "armv6m" "armv6m_soft_nofp" "-mfloat-abi=soft -march=armv6m"
+set(library_variants)
+
+macro(add_library_variants parent_dir_name libc_type)
```
statham-arm, line 347: Perhaps make this a function, so that the large number of internal variables it defines will be confined to the function's own scope and vanish again after the call? You'd have to copy `library_variants` back out into `PARENT_SCOPE` at the end, but I think that's a good tradeoff.

mplatings: Done.

statham-arm (approved):

PR #164, opened 2022-12-12
Build compiler-rt without arch in its filename

mplatings:
This changes the filename from libclang_rt.builtins-<ARCH>.a to simply libclang_rt.builtins.a.


amilendra (approved):
LGTM

PR #165, opened 2022-12-14
Allow using a meson executable not in path

mplatings:
If CMake is run with -DMESON_EXECUTABLE=/path/to/meson then that path should be used.


statham-arm (approved):
It looks as if this also fixes the case in which `find_program` returned a pathname that was not the same one you'd get by searching `$PATH` for `meson`? (Can it do that?)

mplatings:
> It looks as if this also fixes the case in which `find_program` returned a pathname that was not the same one you'd get by searching `$PATH` for `meson`? (Can it do that?)

That's true. find_program does do more than simply looking through PATH

PR #166, opened 2022-12-28
Work around "unable to move location counter backward for: .tbss_space"

mplatings:
The error began with picolibc change
98486a5bbd4f76d45f311b0d950a116c08503938.

The consequence of this workaround is that thread local storage is likely broken.


mplatings:
It's a quiet time of year and the change is very small so merging without review.

PR #169, opened 2023-01-03
Add an AArch64 sample.

voltur01:


miyuki (approved):

mplatings:
@voltur01 to ensure the sample doesn't bitrot I recommend that you add it to the list of samples that are tested: https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/blob/main/CMakeLists.txt#L768

voltur01:
Great point, thanks @mplatings, addressed in https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/pull/170

PR #173, opened 2023-01-17
Infrastructure to support 32-bit A-class targets.

statham-arm:
At the moment, if you try to add a library variant targeting armv7a, the compiler-rt sub-build will silently not build any actual builtins library, because when compiler-rt is given a triple like 'armv7a-none-eabi', it doesn't find 'armv7a' in its list of recognised Arm variants, so it doesn't enable any Arm-related support functions.

A workaround is to put plain 'arm' in the target triple, and specify the arch version using -march=armv7a. However, _that_ doesn't work for v6-M, because compiler-rt needs to include a whole different set of assembly files for that, and if you don't put 'armv6m' in the target triple, it won't notice that it needs to make that change.

So in this change I arrange to only include the specific arch version in the target triple if it's absolutely necessary for something to work, which as far as I know currently means 'only if it's armv6m'. For everything else, the target triple just says 'arm'.

As far as I know that change is only needed for compiler-rt. But it doesn't seem to have done any harm to put it in everywhere.

(Another way to fix this would be to ensure every possible architecture version appears in compiler-rt's list, but that would involve a lot more ongoing work.)

Also, compiler-rt's method of detecting the hard float ABI seems to revolve around being passed 'armhf' as part of the triple. That seemed awkward, so instead I've manually #defined the same symbol that the autodetection would have ended up setting.

There should be no functional change in this commit for any existing variant that we build. But it removes an obstacle to adding v7-A (or v8-A, etc) variants in the future.


CMakeLists.txt
```diff
@@ -431,7 +431,11 @@ while(library_variants_copy)
         set(target "aarch64-none-elf")
         set(cpu_family aarch64)
     else()
-        set(target "${target_arch}-none-eabi")
+        if(target_arch STREQUAL "armv6m")
```
mplatings, line 434: A brief comment here please explaining why this logic is needed

CMakeLists.txt
```diff
@@ -501,6 +505,11 @@ while(library_variants_copy)
 
     set(runtimes_flags "${common_flags} -ffunction-sections -fdata-sections -fno-ident --sysroot ${install_directory}")
 
+    set(compiler_rt_extra_flags)
+    if(flags MATCHES "-mfloat-abi=hard")
+        set(compiler_rt_extra_flags "${compiler_rt_extra_flags} -DCOMPILER_RT_ARMHF_TARGET")
```
mplatings, line 510: A brief comment please explaining what this does, ideally with a link to docs for COMPILER_RT_ARMHF_TARGET

CMakeLists.txt
```diff
@@ -501,6 +505,11 @@ while(library_variants_copy)
 
     set(runtimes_flags "${common_flags} -ffunction-sections -fdata-sections -fno-ident --sysroot ${install_directory}")
 
+    set(compiler_rt_extra_flags)
```
mplatings, line 508: It would be slightly neater to set this to `${runtimes_flags}`
This would simplify the ExternalProject_Add statement a little.

mplatings (changes requested):

CMakeLists.txt
statham-arm: Thanks. In the course of writing the comment explaining why I couldn't do this a nicer way, I found a nicer way :-)

CMakeLists.txt
statham-arm: Done (although I'm not that good at 'brief')

CMakeLists.txt
statham-arm: Now not required, since I've removed that entire mechanism.

mplatings (approved):

PR #174, opened 2023-01-20
Add UBSan sample

voltur01:


samples/src/cpp-baremetal-semihosting-ubsan/libubsan.cpp
```diff
@@ -0,0 +1,54 @@
...
+    HANDLER(invalid_objc_cast, "invalid-objc-cast")
+    HANDLER(function_type_mismatch, "function-type-mismatch")
+    HANDLER(implicit_conversion, "implicit-conversion")
+    HANDLER(nonnull_arg, "nonnull-arg")
+    HANDLER(nonnull_return, "nonnull-return")
+    HANDLER(nullability_arg, "nullability-arg")
+    HANDLER(nullability_return, "nullability-return")
+    HANDLER(pointer_overflow, "pointer-overflow")
+    HANDLER(cfi_check_fail, "cfi-check-fail")
+}
```
mplatings, line 54: terminating newline

samples/src/cpp-baremetal-semihosting-ubsan/libubsan.cpp
```diff
@@ -0,0 +1,54 @@
+extern "C" {
+    #include <stdio.h>
```
mplatings, line 2: Having includes in an `extern "C"` block seems like something to avoid.

mplatings (changes requested):
Can this be added to the list of samples in CMakeLists.txt, to ensure it doesn't bitrot?

voltur01:
Thank you for the review: I fixed the year and added to the list of samples as in the other sample; also updated libubsan.cpp as per your suggestions.

mplatings (approved):

PR #175, opened 2023-01-20
Add CFI sample

voltur01:


samples/src/cpp-baremetal-semihosting-cfi/Makefile
```diff
@@ -0,0 +1,43 @@
+#
+# Copyright (c) 2021, Arm Limited and affiliates.
```
mplatings, line 2: 2023

mplatings (changes requested):
Can this be added to the list of samples in CMakeLists.txt, to ensure it doesn't bitrot?

voltur01:
Thanks for the review: I fixed the year and added to the list of samples.

mplatings (approved):

PR #177, opened 2023-01-25
Also build toolchains for Armv4T, Armv5TE and Armv6

stuij:
LLD and compiler-rt support for these has recently been added to LLVM.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -409,6 +409,8 @@ add_library_variants(aarch64-none-elf picolibc
     aarch64         ""                  "-march=armv8-a"                                        "-M raspi3b"
 )
 add_library_variants(arm-none-eabi picolibc
+    armv4t          ""                  "-march=armv4t"                                         "-M musicpal -cpu arm926"
+    armv5te         ""                  "-march=armv5te"                                        "-M musicpal -cpu arm926"
     armv6m          soft_nofp           "-mfloat-abi=soft -march=armv6m"                        "-M microbit"
     armv7em         hard_fpv4_sp_d16    "-mfloat-abi=hard -march=armv7em -mfpu=fpv4-sp-d16"     "-M mps2-an386 -cpu cortex-m4"
     armv7em         hard_fpv5_d16       "-mfloat-abi=hard -march=armv7em -mfpu=fpv5-d16"        "-M mps2-an500 -cpu cortex-m7"
@@ -437,11 +439,11 @@ while(library_variants_copy)
         # system doesn't recognize every possible Arm architecture
         # version. So mostly we just say 'arm' and control the arch
         # version via -march=armv7m (or whatever).
-        if(target_arch STREQUAL "armv6m")
-            # An exception is armv6m, which compiler-rt expects to see
-            # in the triple because that's where it looks to decide
-            # whether to use the v6-M specific assembly sources.
-            set(target "armv6m-none-eabi")
+        # Exceptions are architectures pre-armv7, which compiler-rt expects to
+        # see in the triple because that's where it looks to decide whether to
+        # use specific assembly sources.
+        if(target_arch MATCHES "^armv[4-6]")
+            set(target "${target_arch}-none-eabi")
         elseif(flags MATCHES "-mfloat-abi=hard")
             # Also, compiler-rt looks in the ABI component of the
             # triple to decide whether to use the hard float ABI.
@@ -581,8 +583,7 @@ while(library_variants_copy)
             )
             add_dependencies(check-picolibc_${variant} check-picolibc_${variant}-${test})
         endfunction()
-
-        if(${variant} MATCHES "^armv8" OR ${variant} MATCHES "^aarch")
+        if(target_arch MATCHES "^(armv8|aarch64)")
             message("Picolibc tests disabled for ${variant}")
         else()
             execute_process(COMMAND ${CMAKE_COMMAND} -E make_directory picolibc_tests_${variant})
--- a/README.md
+++ b/README.md
@@ -23,6 +23,9 @@ embedded and realtime operating systems.
 - Armv7E-M
 - Armv8-M Mainline
 - Armv8.1-M Mainline
+- Armv4T (experimental)
+- Armv5TE (experimental)
+- Armv6 (experimental, using the Armv5TE library variant)
 - AArch64 armv8.0 (experimental)
 
 ## C++ support
--- a/samples/ldscripts/armv4t.ld
+++ b/samples/ldscripts/armv4t.ld
@@ -1,0 +1,7 @@
+__flash =      0x00010000;
+__flash_size = 0x00400000;
+__ram =        0x00500000;
+__ram_size   = 0x00200000;
+__stack_size = 4k;
+
+INCLUDE picolibc.ld
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 - Support for locales and input/output streams
+- Experimental support for Armv4T and Armv5TE architectures
 
 ### Changed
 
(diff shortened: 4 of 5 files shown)


CMakeLists.txt
```diff
@@ -409,6 +409,9 @@ add_library_variants(aarch64-none-elf picolibc
     aarch64         ""                  "-march=armv8-a"
 )
 add_library_variants(arm-none-eabi picolibc
+    armv4t          ""                  "-march=armv4t"
+    armv5te         ""                  "-march=armv5te"
+    armv6           ""                  "-march=armv6"
```
mplatings, line 414: What's the motivation for adding each of these?
There's a cost to building and testing each variant (I personally spend a significant amount of time waiting while this happens) so I want to be sure it benefits someone.

mplatings (commented):
#167 landed this morning so you'll need to update the patch and get the tests passing (or opt out of testing for these architectures if appropriate).

README.md
```diff
@@ -12,12 +12,15 @@ bare-metal LLVM based toolchain targeting Arm based on:
 ## Goal
 
 The goal is to provide an LLVM based bare-metal toolchain that can target the
-Arm architecture family from Armv6-M and newer. The toolchain follows the ABI
+Arm architecture family from Armv4T and onwards. The toolchain follows the ABI
```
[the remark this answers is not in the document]
voltur01: I suggest that we keep the goal as-is. Additional configurations can be added as experimental similarly to aarch64 that we have today.

Provided that there are many experimental configurations we may add a sub-section or separate them visually in some other way.

At least, I would move all experimental configurations to the end of the list, i.e. do not start the list with experimental ones, but focus on the configurations from the goal statement.

samples/ldscripts/armv4t.ld
```diff
@@ -0,0 +1,7 @@
+__flash =      0x00000000;
```
voltur01, line 1: These appear to be the same in all 3 new ld files - can this be generalized to provide one file? or do you expect them to be different in the future?

BTW, does this specify a specific board? Maybe then the file can be named after the board like we have for micro:bit and Raspberry Pi.

voltur01 (changes requested):

samples/ldscripts/armv4t.ld
stuij: I've followed the convention from the picolibc test patch which landed today. Currently there are 9 other identical files in that folder. @pwprzybyla and @mplatings would be in a better position to answer.

README.md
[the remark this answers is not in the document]
stuij: ack

samples/ldscripts/armv4t.ld
voltur01: Hm, now I see them, though the question is do they belong more to the tests folder, than samples? that way it would be clear that they follow the config files naming - test config for every library config.

CMakeLists.txt
mplatings: If `v4t` covers all those use cases then could you reduce it to adding just that one? It sounds like v5te and v6 would be optimisations.

samples/ldscripts/armv4t.ld
mplatings: @voltur01 I'm fine if you want to move the files around in a different change. For now, I think where Ties has put the files is appropriate.

mplatings:
> we should opt out of testing for v4t as it appears no v4t cpu is available in qemu

You can add that to this line:

```
        if(${variant} MATCHES "^armv8" OR ${variant} MATCHES "^aarch")
            message("Picolibc tests disabled for ${variant}")
```

samples/ldscripts/armv4t.ld
voltur01: I agree that it is a possible refactoring on the testing side, not related to this patch, so OK with me.

CMakeLists.txt
stuij: Between 5te and 6 I agree that the difference would be minimal. But between v4t and v5te there have been enough improvements in how arm-thumb interworking is done (every such function call would be affected) that I think it makes sense to have the two variants. Also because people in embedded are quite sensitive about code size.

I'm a little bit hesitant about the optics of having people building a v6 app with a v5 library. They should be clued up enough about the compatibility story that they feel secure enough to use the v5 variant. Which I don't think is a given. I guess we can give a note of it in the documentation, and as we don't yet have a real-world use case, it's probably fair enough to omit it.

mplatings: OK, v4t and v5te it is. v6 can always be added later if a need is identified.
Once [multilib](https://discourse.llvm.org/t/rfc-multilib/67494) eventually lands people won't have to think about which library they should use :)

samples/ldscripts/armv6.ld
```diff
@@ -0,0 +1,7 @@
+__flash =      0x00000000;
```
mplatings, line 1: Remove this file now it's not used

CMakeLists.txt
```diff
@@ -437,10 +439,16 @@ while(library_variants_copy)
...
+        # Exceptions are architectures pre-armv7, which compiler-rt expects to
+        # see in the triple because that's where it looks to decide whether to
+        # use specific assembly sources.
+        if(target_arch STREQUAL "armv4t")
+            set(target "armv4t-none-eabi")
+        elseif(target_arch STREQUAL "armv5te")
+            set(target "armv5te-none-eabi")
+        elseif(target_arch STREQUAL "armv6")
+            set(target "armv6-none-eabi")
+        elseif(target_arch STREQUAL "armv6m")
```
mplatings, line 451:
```suggestion
        if(target_arch MATCHES "^armv[4-6]")
            set(target "${target_arch}-none-eabi")
```

CMakeLists.txt
```diff
@@ -582,7 +590,7 @@ while(library_variants_copy)
             add_dependencies(check-picolibc_${variant} check-picolibc_${variant}-${test})
         endfunction()
 
-        if(${variant} MATCHES "^armv8" OR ${variant} MATCHES "^aarch")
+        if(${variant} MATCHES "^armv4t" OR ${variant} MATCHES "^armv8" OR ${variant} MATCHES "^aarch")
```
mplatings, line 593:
```suggestion
        if(target_arch MATCHES "^(armv4t|armv8|aarch64)")
```

mplatings (changes requested):
I gave it a try and I get errors like these:
```
[2993/3302] TEST: armv5te test-efcvt
FAILED: CMakeFiles/check-picolibc_armv5te-test-efcvt /work/LLVM-embedded-toolchain-for-Arm/build-docker/CMakeFiles/check-picolibc_armv5te-test-efcvt 
cd /work/LLVM-embedded-toolchain-for-Arm/build-docker/picolibc_tests_armv5te && qemu-system-arm -M mps2-an386 -cpu arm946 -semihosting -nographic -kernel check-picolibc_armv5te_build_test-efcvt.out
qemu-system-arm: This board can only be used with CPU cortex-m4-arm-cpu
```
This is with QEMU 6.0.0

CMakeLists.txt
stuij: Right, unfortunately I'm getting a linker error on all platforms when compiling the picolibc tests (on otherwise clean repo) which I still need to get to the bottom of, so haven't been able to test this yet. Turns out machine <> cpu compatibility checking in QEMU didn't work like I expected. Looking closer, there's a Canon board that works with v5te (as well as the generic board I expect), but I'll test it first before patching the patch.

stuij: fixed v5te picolibc tests. was an issue with setting up the men regions for appropriate v5te board.

stuij:
The last patch should have fixed all the outstanding issues.

One thing I wasn't sure of: we don't have a QEMU board that supports v4t (yet). But we can test the code on a v5te board. On the one hand it seems a bit misleading to test on a newer (but compatible) arch. On the other hand we would test the generated code, which is better than not testing at all. I'm leaning towards enabling v4t testing on a v5te board.

Thoughts?

mplatings:
I strongly prefer enabling testing on a compatible board

stuij:
> I strongly prefer enabling testing on a compatible board

done!

mplatings (approved):

mplatings (changes requested):
Oops one more change please - this is a noteworthy enhancement so please add it in CHANGELOG.md

mplatings (approved):

PR #182, opened 2023-02-13
Support running C++ runtime libraries tests

miyuki:


test-support/lit-exec-qemu.py
```diff
@@ -0,0 +1,64 @@
...
+import argparse
+import subprocess
+import pathlib
+
+# This script requires Python 3.6 or later
+assert sys.version_info >= (3, 6)
+
+
+def run(args):
+    """Execute the program using runner.pl and return the subprocess return
```
statham-arm, line 20: Nit: surely 'using qemu'

test-support/lit-exec-qemu.py
```diff
@@ -0,0 +1,64 @@
...
+    assert command
+    if command[0] == '--':
+        command = command[1:]
+    image = command[0]
+    image_args = command[1:]
+    qemu_params = args.qemu_params.split(':')
+    qemu_cmd = ([args.qemu_command] + qemu_params +
+        ['-semihosting', '-nographic', '-kernel', image])
+    if image_args:
+        image_args_concat = ','.join(['arg=' + arg for arg in image_args])
```
statham-arm, line 32: What happens if there's a comma in one of the arguments? If QEMU provides an escaping mechanism then we should use it. If it doesn't, then giving an error is preferable to silently doing something confusing here.

statham-arm (approved):

PR #188, opened 2023-02-27
Windows compatibility changes

mplatings:
* Update picolibc patch based on feedback from https://github.com/picolibc/picolibc/pull/435
* Avoid calling shell script when building picolibc
* Enable building runtimes on Windows. Also add CMAKE_EXECUTABLE_SUFFIX exhaustively for all executable paths.


miyuki (approved):
LGTM

PR #191, opened 2023-03-02
Picolibc tests enabled for armv8*

pwprzybyla:


mplatings:
Great to see progress on this.
* Can you explain in the commit message what you're changing and why?
* Please avoid unnecessary whitespace changes.

pwprzybyla:
While investigating HardFault's with picolibc tests for armv8* variants. I have discovered that leaving default picolibc values in terms of RAM resolves most of the HardFaults with tests. Only few platforms require increased flash space.

mplatings:
I gave this a try and check-picolibc_aarch64 seems pretty broken

pwprzybyla:
> 

Aarch64 is disabled again as it was. All tests in bundle check-picolibc pass.

PR #192, opened 2023-03-07
Allow copying libraries from another build

mplatings:
Building the libraries on Windows can take hours, compared to minutes on Linux. Therefore this change allows copying libraries from another build.

The new PREBUILT_TARGET_LIBRARIES CMake option causes the building of libraries to be skipped, with the assumption that the libraries will be added in the location in which they would normally be built.

The copy_target_libraries.py script can be used to copy the libraries from a previously built package.

### DIFF
--- a/cmake/copy_target_libraries.py
+++ b/cmake/copy_target_libraries.py
@@ -1,0 +1,74 @@
+#!/usr/bin/env python3
+
+"""
+Script to copy target libraries into the build tree.
+Building libraries can take a very long time on some platforms so
+building them on another platform and copying them in can be a big
+time saver.
+"""
+
+import argparse
+import glob
+import os
+import shutil
+import tarfile
+import tempfile
+
+
+def main():
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--distribution-tgz",
+        required=True,
+        help="""Copy from this LLVM Embedded Toolchain for Arm distribution
+        tar.gz file. This is a glob to make things easier on Windows.""",
+    )
+    parser.add_argument(
+        "--build-dir",
+        required=True,
+        help="The build root directory to copy into",
+    )
+    args = parser.parse_args()
+
+    # Find the distribution. This is a glob because scripts may not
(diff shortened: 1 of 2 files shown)


cmake/copy_target_libraries.py
```diff
@@ -0,0 +1,74 @@
...
+import tempfile
+
+
+def main():
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--distribution-tgz",
+        required=True,
+        help="""Copy from this LLVM Embedded Toolchain for Arm distribution
+        tar.gz file. This a glob to make things easier on Windows.""",
```
statham-arm, line 24: "This a glob" no verb.

cmake/copy_target_libraries.py
```diff
@@ -0,0 +1,74 @@
...
+        "--build-dir",
+        required=True,
+        help="The build root directory to copy into",
+    )
+    args = parser.parse_args()
+
+    # Find the distribution. This is a glob because scripts may not
+    # know the version number and we can't rely on the Windows shell to
+    # do it.
+    for distribution_tgz in glob.glob(args.distribution_tgz):
```
statham-arm, line 36: I see that on Windows it makes sense to make this a glob, but on Unix, this reduces generality, because some legal filenames become difficult to use. I think it would be better to only do the globbing on Windows, where it doesn't reduce generality because of the much tighter restrictions on valid filenames.

statham-arm (changes requested):

cmake/copy_target_libraries.py
statham-arm: I agree this script is more likely to be run on Windows – but for just that reason, surely it's the more complex Windows-specific code that _will_ be tested! If anything, the untested code branch likely to fail is the 'don't bother globbing this, we're on Unix' version.

Still, perhaps you're right for a different reason: there are enough unquoted pathnames in the build scripts of all the contained packages that it's already likely not feasible to build this whole system inside a directory that contains strange characters. I've noticed that even in LLVM itself.

statham-arm (approved):

PR #198, opened 2023-03-15
Enable Picolibc tests: long_double, math_errhandling, timegm on suita…

pwprzybyla:
…ble platforms

More picolibc tests are enabled now.
Additionally failing tests were explicitly disabled for each variant. One can see now how stable given test/variant is.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -480,14 +480,36 @@ function(add_picolibc directory variant target_triple flags qemu_params variant_
         tls
         ungetc
     )
-    #remove tests that fail at least at one platform
-    list(REMOVE_ITEM picolibc_tests abort constructor-skip long_double malloc malloc_stress math_errhandling test-except timegm tls)
     # remove tests failing on a given platform
+    if(variant STREQUAL "armv6m_soft_nofp")
+        list(REMOVE_ITEM picolibc_tests abort constructor-skip test-except tls)
+    endif()
+    if(variant STREQUAL "armv7m_soft_nofp")
+        list(REMOVE_ITEM picolibc_tests abort constructor-skip test-except tls)
+    endif()
+    if(variant STREQUAL "armv7em_soft_nofp")
+        list(REMOVE_ITEM picolibc_tests abort constructor-skip test-except tls)
+    endif()
+    if(variant STREQUAL "armv7em_hard_fpv4_sp_d16")
+        list(REMOVE_ITEM picolibc_tests abort constructor-skip math_errhandling test-except tls)
+    endif()
+    if(variant STREQUAL "armv7em_hard_fpv5_d16")
+        list(REMOVE_ITEM picolibc_tests abort constructor-skip math_errhandling test-except tls)
+    endif()
+    if(variant STREQUAL "armv8m.main_soft_nofp")
+        list(REMOVE_ITEM picolibc_tests abort constructor-skip long_double math_errhandling malloc_stress test-except timegm tls)
+    endif()
     if(variant STREQUAL "armv8m.main_hard_fp")
-        list(REMOVE_ITEM picolibc_tests complex-funcs fenv math-funcs printf_scanf rand test-efcvt test-strtod)
+        list(REMOVE_ITEM picolibc_tests abort constructor-skip complex-funcs fenv long_double math_errhandling math-funcs malloc_stress printf_scanf rand test-efcvt test-except test-strtod timegm tls)
+    endif()
+    if(variant STREQUAL "armv8.1m.main_soft_nofp_nomve")
+        list(REMOVE_ITEM picolibc_tests abort constructor-skip long_double math_errhandling malloc_stress test-except timegm tls)
+    endif()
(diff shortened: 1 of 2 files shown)


mplatings (changes requested):
Thanks for this, and well done for getting it working.

I like that tests are disabled per-variant, it does make it clearer where the gaps are.

I'd prefer to avoid changes to picolibc unless absolutely necessary. Is there a model that's larger than microbit that we could use? I see that picolibc's `run-arm` uses `-machine virtm -cpu cortex-m0` for its Armv6-M testing so does it make sense to do the same here?

patches/picolibc.patch
```diff
@@ -1,11 +1,11 @@
 diff --git a/meson.build b/meson.build
-index 62384e63c..a242da5bf 100644
```
mplatings, line 2: I don't see any significant changes to this file so please remove it from the commit.

CMakeLists.txt
```diff
@@ -817,7 +838,7 @@ add_library_variants(
     aarch64         ""                  "-march=armv8-a"                                        "-M raspi3b"
     armv4t          ""                  "-march=armv4t"                                         "-M musicpal -cpu arm926"
     armv5te         ""                  "-march=armv5te"                                        "-M musicpal -cpu arm926"
-    armv6m          soft_nofp           "-mfloat-abi=soft -march=armv6m"                        "-M microbit"
+    armv6m          soft_nofp           "-mfloat-abi=soft -march=armv6m"                        "-M mps2-an385,accel=tcg"
```
mplatings, line 841: Comment please to explain why `accel` is needed and what `tcg` means

CMakeLists.txt
```diff
@@ -519,11 +541,10 @@ function(add_picolibc directory variant target_triple flags qemu_params variant_
             picolibc_test(${test})
         endforeach()
         # Test disabled due to duplicated symbol name div in test file and in picolibc
-        # picolibc_test(rounding-mode ${picolibc_SOURCE_DIR}/test/rounding-mode-sub.c)
+        # picolibc_test(rounding-mode)
         # Test disabled due to fail on armv8.1m.main_hard_nofp_mve
-        # picolibc_test(try-ilp32 ${picolibc_SOURCE_DIR}/test/try-ilp32-sub.c)
+        # picolibc_test(try-ilp32)
         add_dependencies(check-picolibc check-picolibc_${variant})
-
```
mplatings, line 526: Avoid whitespace changes on lines that aren't connected to the main change you're making

CMakeLists.txt
```diff
@@ -480,14 +480,36 @@ function(add_picolibc directory variant target_triple flags qemu_params variant_
...
+    endif()
+    if(variant STREQUAL "armv7em_hard_fpv4_sp_d16")
+        list(REMOVE_ITEM picolibc_tests abort constructor-skip math_errhandling test-except tls)
+    endif()
+    if(variant STREQUAL "armv7em_hard_fpv5_d16")
+        list(REMOVE_ITEM picolibc_tests abort constructor-skip math_errhandling test-except tls)
+    endif()
+    if(variant STREQUAL "armv6m_soft_nofp")
+        list(REMOVE_ITEM picolibc_tests abort constructor-skip test-except tls)
     endif()
```
mplatings, line 513: Nit: please sort these in a way that is consistent with the list passed to add_library_variants()

samples/ldscripts/armv6m_soft_nofp.ld
```diff
@@ -1,7 +1,7 @@
-__flash = 0x00000000;   /* starting address of flash */
```
mplatings, line 1: This file is actually part of a sample used for educational purposes so please restore the comments

mplatings (changes requested):

pwprzybyla (commented):
Change requests are addressed now.

CMakeLists.txt
pwprzybyla: Actually accel tcg is not necessary to have tests working so it's removed now.

CMakeLists.txt
pwprzybyla: Done.

CMakeLists.txt
pwprzybyla: It's sorted in same way as list of variants is sorted.

samples/ldscripts/armv6m_soft_nofp.ld
pwprzybyla: Was not aware bout that. Comments are back.

samples/ldscripts/armv6m_soft_nofp.ld
mplatings, line 1: Since we're forced to use a Cortex-M3 model for Armv6-M tests, the filepath of this file is misleading.

A simple solution is to move the ldscripts needed for these tests under the `tests` directory. For example `tests/ldscripts`. If a file is needed for *both* samples and tests then leave a copy of it under `samples`.

Another possibility is that you rename the ldscripts according to the machine name. For example `mps2-an385.ld`.

mplatings:
I think this was replaced by #211

Issue #197, opened 2023-03-12
Include profiling lib?


rgrr:
I'm trying to do profiling on the target with clang.  Unfortunately the corresponding symbols __llvm_profile* are not included in clang_rt.

I have tried to include it via `-DCOMPILER_RT_BUILD_PROFILE=ON` in CMakeLists.txt:~600 but had no success.  On the internet there is almost no information about how to build this part correctly.

Do you have any pointer or better could you please include it in the toolchain package?

voltur01:
Hi Hardy,

Indeed, the profiling runtime is not included since it does not support building for bare-metal targets now. We are interested in enabling it, but we expect to be able to look into it at some later stage.

As a workaround, what should be possible it to provide own implementation for the profiling runtime functions required in a particular use case, however this would involve reverse engineering of what these runtime functions are expected to do and re-implementing them.

smithp35:
I have been able to write an implementation of the profiling functions for code-coverage in LLVM 13.0.0 although sadly the profiling format changes between LLVM releases and my runtime was written for a different linker and C-library so it would not work without modifications.

The bare-metal runtime for coverage is based on the code in compiler-rt/lib/profile/InstrProfilingPlatformOther.c I added my own __llvm_profile_dump() function that was registered as an atexit handler by `__llvm_profile_register_function()`. This essentially writes a file with the profiling header and then writes out the various counter sections to a file. An alternative method would be to call the __llvm_profile_dump() directly either at the end of main, or immediately after it.

If I get the time I'll try and adapt this for LLVM 16. May take me some time to do this though.

The main difficulty for bare-metal profiling upstream is that there is no universal way of extracting the counters.

rgrr:
Hello Volodymyr & Peter,

some time ago I implemented the profiling output for clang13 via debug console output too.  But as Peter wrote, those implementations are compiler dependent and I'm stuck at clang13 as well (also I did separation of the constant and variable data, because the profiling stuff contains a lot of constant data (sections __llvm_covmap, __llvm_covfun) which should not go into the target, but perhaps this has changed in the meantime).

I'm wondering if there could be a more generic approach via semihosting.  Semihosting just for transmission of the profiling data after execution, for nothing else.  This way, the regular profiling dump code should be usable.

Currently I'm trying to change the build operations of this project to output the profiling library (libclang_rt.profile*.a ?) to further experiment.  But til now without luck.

Any pointers for doing this?

[...]

PR #204, opened 2023-03-21
Profiling contribution

rgrr:
Referencing discussion in #197, this is my proposal.

This PR adds a contrib subdirectory structure to the project which consists currently of one feature: profiling with semihosting output.

The profiling library is created for every enabled target and placed in each $(SYSROOT)/lib.

Feature can be enabled on the cmake command line via -DLLVM_TOOLCHAIN_CONTRIB_PROFILE=ON, default is OFF.

Detailed information in README.md in the PR.

Note that this is not the final version (because the root CMakeLists.txt contains some unwanted changes).  This PR should be considered for discussion and I'm awaiting feedback.


rgrr:
Hi Michael, I understand your point & see it the same way.  That's the reason why this additional feature resides in contrib and is not built by default.

If you have a look at the CMakeLists.txt of the profile library, you will notice that most of the profiling code is taken from the clang sources.  This is intentional to keep the effects of impending upstream changes to a minimum.

The core functions of the semihost writing thus are still taken from the clang sources and the major code contribution which could be upstream dependant is `fileWriter()`.

```C
static uint32_t fileWriter(ProfDataWriter *This, ProfDataIOVec *IOVecs, uint32_t NumIOVecs)
{
    uint32_t I;
    int fd = (int)This->WriterCtx;
    char Zeroes[sizeof(uint64_t)] = {0};
    for (I = 0; I < NumIOVecs; I++)
    {
        size_t Length = IOVecs[I].ElmSize * IOVecs[I].NumElm;

        if (IOVecs[I].Data)
        {
            if (SEMI_WRITE(fd, IOVecs[I].Data, Length) != Length)
                return 1;
        }
        else if (IOVecs[I].UseZeroPadding)
        {
            while (Length > 0)
            {
                size_t PartialWriteLen = (sizeof(uint64_t) > Length) ? Length : sizeof(uint64_t);
                if (SEMI_WRITE(fd, Zeroes, PartialWriteLen) != PartialWriteLen)
                {
                    return 1;
                }
                Length -= PartialWriteLen;
            }
        }
    }
    return 0;
}
```

This code has been borrowed from InstrProfilingFile.c.  If you check its blame: https://github.com/llvm/llvm-project/blame/main/compiler-rt/lib/profile/InstrProfilingFile.c you can see that this block has been stable for at least 4-8 years.  And to be honest I do not expect there big changes in the future - except if they change the concept at all.

The other code is more or less semihosting stuff and file open & close and I doubt that this is very much version dependant or will change from version to version.

As Peter already pointed out, injecting the code upstream will be very difficult, because there is no standard way for data transfer.  My proposal is based on Arm(!) semihosting(!), others prefer perhaps output via debug console.  If I remember correctly TI does dumping inside the debugger (but not with semihosting).  So there are many approaches to put under one hat.  And Apple has completely different ideas as one can see in InstrProfilingFile.c.

To my opinion there is no chance to get this upstream.

To sum up: I tried to recycle most of the required code from upstream to minimize danger of breakage.  And chances to get this solution upstream is almost zero because it is specialized to Arm and semihosting.  Which is one of the specials of the toolchain build with this repo ;-)

rgrr:
> > To my opinion there is no chance to get this upstream.
> 
> Let's not pre-judge that. From the [thread Peter linked to](https://lists.llvm.org/pipermail/llvm-dev/2017-September/117156.html) there is an appetite for this in llvm-project.

Yes, there is an appetite for a solution, unfortunately for a generic solution.  But my solution is not generic, in fact it is very limited and special.  This would add another special case to their profile handling.  And I doubt that they will be happy about that.

And this is also not end of story for this "embedded toolchain", because I started trying to compile the existing profiling library.  Without success.  So there will be some work to be done.

Here you have a ready solution for an (I admit: not very often) asked question.  The impact on the actual toolchain is none and the user/developer has to explicitly switch on the feature and usage as well.  He/she even has to compile the toolchain him/herself, so there must be some conscious action before usage. So no harm to standard conditions.

And your before mentioned fear that every change of LLVM impacts on this code can be delegated to me (I promise ;-)).

> I feel the best way forward here is to open a review following the [LLVM Contributions process](https://llvm.org/docs/Contributing.html). In terms of finding reviewers, the people who contributed to the above thread would be a good start.

I cannot share your confidence in that.  I have a pending issue with a clear bug (at least to my opinion) with no actual progress.  And there are several hundred others.  I doubt that a "new" half-brewed feature will be on top of their list.

mplatings:
I understand that your experience with raising a GitHub issue for llvm-project has been off-putting. However commenting on a review requires less time from the LLVM community than taking on a bug fix so they are treated differently. I won't promise it will be fast but with persistence and patience progress can be made. I'm certain you're right that they won't accept your solution as-is, but they can probably point you in the direction of what would be acceptable.

rgrr:
Can I treat this as a refusal? :-(

Ok then... what would you recommend?  Raising an issue or filing a PR directly?

Nevertheless a problem persists: compiler-rt does not build InstrProfiling and I haven't found yet what's the cause.

mplatings:
https://llvm.org/docs/Contributing.html#how-to-submit-a-patch

voltur01:
Hi,

May I suggest to put this forward as an agenda item for the LLVM community call about embedded toolchains, next one is next Thursday 30th of March, see https://discourse.llvm.org/t/llvm-embedded-toolchains-working-group-sync-up/63270

The topic of runtimes for profiling and sanitizers was flagged as something the embedded community is interested in and was on the list of topis to follow up anyway.

Hopefully, we can get an agreement on the overall direction how to tackle it upstream.

rgrr:
Hello Volodymyr,

I think this is a good idea.  Yesterday I made some trials with Phabricator and I must say that they created some hurdles for "newcomers".

Clarification if such contributions are welcome would make it a little bit more motivating to go thru the process.

And if the result is, that such contributions like mine are too specific and should remain in projects like this one, wouldn't be too bad for me as well ;-)

voltur01:
FYI meeting minutes of the discussion https://discourse.llvm.org/t/llvm-embedded-toolchains-working-group-sync-up/63270/25

In short, there is interest and even some patches around - need to coordinate and start actual improvements, contributions are welcome and will be supported. Will be discussed in the coming EuroLLVM as well.

rgrr:
> FYI meeting minutes of the discussion https://discourse.llvm.org/t/llvm-embedded-toolchains-working-group-sync-up/63270/25
> 
> In short, there is interest and even some patches around - need to coordinate and start actual improvements, contributions are welcome and will be supported. Will be discussed in the coming EuroLLVM as well.

Ok, that sounds good.  I will give it a try and check Phabricator again.

rgrr:
Setting up the Phabricator process for the first time is very confusing (TMO) and the steps following are not very clear.

Current hurdle is, that Bill Wendling (CODE_OWNERS.txt of compiler-rt) does not seem to exist any more.

So I'm stuck there.  Any idea what to specify as reviewer for profiling?

stuij:
Or as one of our colleagues tends to say: using git blame, look at Phabricator reviewers of patches that have gone in recently in that code area. The strike rate is probably bigger.

rgrr:
I'm starting to think that I'm too stupid to make Phabricator work for me.  I've entered brad@comstyle.com as a reviewer.  Not happy with it.  I provided wyehia@ca.ibm.com, still not happy.
So I'm still stuck with submitting my additional module :-/

mplatings:
Putting those emails into the reviews.llvm.org search box and clicking around a bit suggests you're looking for the user names `brad` and `w2yehia`. You can add them both as reviewers and see who responds.

mplatings:
Discord is a good place to ask for LLVM beginner help: https://discord.gg/xS7Z362

voltur01:
Hi @rgrr, any progress on this topic? We have the next LLVM Embedded Toolchains WG call this week, so I can add to the agenda, if there are any updates.

PR #205, opened 2023-03-21
Add guidance on how & why to contribute

mplatings:


docs/contributing.md
```diff
@@ -1,8 +1,37 @@
...
+
 Contributions are accepted under Apache-2.0. Only submit contributions where
 you have authored all of the code.
 
+This project follows the conventional
+[GitHub pull request](https://docs.github.com/en/pull-requests) flow.
+
+## Testing
+
+Please ensure your change doesn't break tests. (The project doesn't yet have
```
smithp35, line 27: Suggesst "doesn't break the tests"

docs/contributing.md
```diff
@@ -1,8 +1,37 @@
...
+GitHub actions to do this). Except for documentation changes, please check that
+this passes:
```
smithp35, line 29: suggest instead of "this passes:" "the following passes:"

smithp35 (approved):
Looks good to me.

I've made a couple of suggestions for the language, but these are entirely subjective, feel free to ignore.

voltur01:
The addition looks good to me, as an extension of this PR or a separate one, we can also extend along the lines of https://llvm.org/docs/Contributing.html, e.g. mention the way to contribute: report an issue or contribute a patch. For reporting issues we should also suggest that upstream issues should be reported in upstream projects.

We can also mention that for bigger changes/new features it may be helpful to have an RFC as a Github issue first.

I can try to suggest some wording in a follow up PR, if you prefer.

mplatings:
Yes let's do a separate PR for that, thanks

PR #206, opened 2023-03-22
Don't strip binaries when creating package

mplatings:
Stripping removes code signatures. Therefore if stripping needs to be done it must be done prior to signing, and not at package creation time.


statham-arm (commented):
The commit message leaves it ambiguous whether you've _actually_ moved the stripping to before the signing, or merely removed the post-signing strip step.

Looking at the patch, I can see it's the latter, but perhaps mention that in the commit message to save the next reader the trouble?

statham-arm (approved):

PR #211, opened 2023-03-28
Move ldscripts to test path

pwprzybyla:
LDSCRIPTS are moved to test path.


CMakeLists.txt
```diff
@@ -518,7 +518,7 @@ function(add_picolibc directory variant target_triple flags qemu_params variant_
         add_custom_command(
                 OUTPUT picolibc_tests_${variant}/check-picolibc_${variant}_build_${test}.out
                 COMMAND ${LLVM_BINARY_DIR}/bin/clang${CMAKE_EXECUTABLE_SUFFIX} --config ${LLVM_BINARY_DIR}/bin/${variant}_semihost.cfg
-                -T ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/samples/ldscripts/${variant}.ld
+                -T ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/tests/ldscripts/picolibc_${variant}.ld
```
mplatings, line 521: It looks like line 555 needs updating in the same way. Maybe others as well

mplatings:
I request that you run `ninja check-llvm-toolchain` with all libraries enabled before putting code for review. For me it gives the error `ld.lld: error: cannot find linker script /work/LLVM-embedded-toolchain-for-Arm/tests/ldscripts/picolibc_armv4t.ld`

pwprzybyla:
Indeed! It was working for me as I explicitly typed LLVM_TOOLCHAIN_LIBRARY_VARIANTS without armv4t and armv5te variants. Should we also support picolibc library for them?

mplatings:
You don't need to go out of your way to add testing for them. My main concern here is that this change breaks the build.

pwprzybyla:
Actually armv4t and armv5te works pretty well with picolibc (I judge it based on tests). So if it's ok I would like to support them same way the other variants are supported.

mplatings (approved):

PR #213, opened 2023-04-03
Freeze llvm-project and picolibc sources.

mplatings:
* Use llvm-project version 16.0.0 tag.
* picolibc hasn't been released since December and has important recent changes so use the latest revision.


simpal01 (approved):
Looks good to me.

PR #214, opened 2023-04-04
don't add "origin/" when checking out a tag

stuij:
In read_repo_version in CMakeLists.txt we look at versions.json to know what version of a repo we should check out. We can specify the version type be either commithash, branch or tag. Currently we add "origin/" to the version string if the type is either branch or tag. However for tag this isn't the right syntax. Removing origin will properly check out the repo by tag.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,11 +203,13 @@ function(read_repo_version output_variable_prefix repo)
     if(tagType STREQUAL "commithash")
         # GIT_SHALLOW doesn't work with commit hashes.
         set(shallow OFF)
-    elseif(tagType STREQUAL "tag" OR tagType STREQUAL "branch")
+    elseif(tagType STREQUAL "branch")
         set(shallow ON)
         # CMake docs recommend that "branch names and tags should
         # generally be specified as remote names"
         set(tag "origin/${tag}")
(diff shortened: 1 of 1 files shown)


stuij:
without this fix we error:
```
[0/7] Performing update step for 'llvmproject-populate'
From https://github.com/llvm/llvm-project
   8ec36e6956..4b2be144e1  main       -> origin/main
error: pathspec 'origin/llvmorg-16.0.0' did not match any file(s) known to git
```

CMakeLists.txt
```diff
@@ -203,11 +203,14 @@ function(read_repo_version output_variable_prefix repo)
...
         set(shallow OFF)
-    elseif(tagType STREQUAL "tag" OR tagType STREQUAL "branch")
+    elseif(tagType STREQUAL "branch")
         set(shallow ON)
         # CMake docs recommend that "branch names and tags should
         # generally be specified as remote names"
         set(tag "origin/${tag}")
+    elseif(tagType STREQUAL "tag")
+        set(shallow ON)
+        set(tag "${tag}")
```
mplatings, line 213: delete this line

stuij: sorry, what line?

mplatings: `        set(tag "${tag}")`

PR #215, opened 2023-04-04
don't add "origin/" when checking out a tag

stuij:
In read_repo_version in CMakeLists.txt we look at versions.json to know what version of a repo we should check out. We can specify the version type be either commithash, branch or tag. Currently we add "origin/" to the version string if the type is either branch or tag. However for tag this isn't the right syntax. Removing origin will properly check out the repo by tag.


stuij:
cherry-picking from main

mplatings (approved):

PR #220, opened 2023-04-11
Reinstate -march=armv8-a for AArch64 library

mplatings:
Without it, `ninja check-llvm-toolchain` hangs.


statham-arm (commented):
Post-commit: LGTM

PR #222, opened 2023-04-11
Update changelog

mplatings:
Moved multilib details to README.md and expanded them.


statham-arm (approved):
Good idea to document the escape hatch of explicit `--sysroot` – if I'd written this patch I think I'd have forgotten to do that.

PR #226, opened 2023-04-12
Update samples README with macOS information

smithp35:
The LLVM Embedded Toolchain for Arm has a macOS release. Update the samples README to claim support and to point the reader to the instructions for installing QEMU.

macOS requires the developer tools to be installed to run the samples, which also installs LLDB. The existing instructions for Linux work for macOS.

### DIFF
--- a/samples/README.md
+++ b/samples/README.md
@@ -13,6 +13,7 @@ Embedded Toolchain for Arm.
 
 The build scripts support three different environments:
 * Linux
+* macOS
 * Windows
 * Windows + [MSYS2](https://www.msys2.org/)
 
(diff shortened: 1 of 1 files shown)


samples/README.md
```diff
@@ -52,7 +57,7 @@ to the ``bin`` directory of the toolchain.
 
 ## Compiling, running and debugging the samples
 
-### Linux and MSYS2
+### Linux macOS and MSYS2
```
mplatings, line 60:
```suggestion
### Linux, macOS and MSYS2
```

samples/README.md
```diff
@@ -13,14 +13,15 @@ Embedded Toolchain for Arm.
 
 The build scripts support three different environments:
```
stuij, line 14: I guess this should now be: "4 different environments"

stuij:
Also, the root README and the build instructions still have macOS as being experimental. But at this point it's perhaps not more experimental than Windows and Linux.

smithp35:
> Also, the root README and the build instructions still have macOS as being experimental. But at this point it's perhaps not more experimental than Windows and Linux.

That's a good point, given the title I think that's a different pull request though. Can do that separately.

stuij:
LGTM

mplatings (approved):

PR #235, opened 2023-04-18
Fix picolibc tests abort and test-except

pwprzybyla:
Tests are fixed by usage of test executor
which substitues return code for picolibc tests
that on success return non zero code

### DIFF
--- a/samples/src/baremetal-uart/hello.c
+++ b/samples/src/baremetal-uart/hello.c
@@ -34,18 +34,26 @@ void uart_init()
   delay(1000);
 }
 
-void uart_send_message(char* message)
+int uart_putc(char ch, FILE* file)
 {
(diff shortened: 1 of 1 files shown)


mplatings (changes requested):
Thanks for this Piotr. I can see that this will work but I think it's the wrong approach in the long term. Can you please first refactor the tests to be run via llvm lit? Then instead of hacking lit-exec-qemu.py to allow non-zero exit codes you can use `not` in the lit test file. See https://github.com/llvm/llvm-project/blob/main/clang/test/Driver/crash-report.cpp#L15 as an example.

If you're not already familiar with lit then you'll need to take some time to familiarise yourself with it. https://llvm.org/docs/TestingGuide.html gives a good overview.

PR #236, opened 2023-04-21
Allow customising which parts of the toolchain to install

mplatings:


smithp35 (commented):
I agree with the logic and can't spot anything obviously wrong. I'm not much of a CMake expert so I can't suggest any improvements.

smithp35 (approved):
Apologies forgot to press the approve button.

PR #237, opened 2023-04-21
Remove dependency on libtinfo.so

mplatings:
Users without sudo permission won't be able to `apt install libtinfo5` so the need to do so can be a significant stumbling block. The benefit of libtinfo is a relatively small usability perk so remove it.


smithp35 (approved):
Looks good to me. I think accessibility of the toolchain is more important than the coloured text.

PR #238, opened 2023-04-24
Add compiler-rt testing

dcandler:
This adds testing for compiler-rt, and begins to collect runtime testing under the target check-llvm-toolchain-runtimes, which can also be used to run testing on specific variants.

There are currently a few tests which fail, which require upstream changes to fix the tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -380,9 +380,19 @@ add_custom_target(llvm-toolchain ALL)
 add_custom_target(llvm-toolchain-runtimes)
 
 # Groups all C++ runtime libraries tests
+add_custom_target(check-llvm-toolchain-runtimes)
+add_custom_target(check-compiler-rt)
 add_custom_target(check-cxxabi)
 add_custom_target(check-unwind)
 add_custom_target(check-cxx)
+
+add_dependencies(
+    check-llvm-toolchain-runtimes
+    check-compiler-rt
+    check-cxxabi
+    check-unwind
+    check-cxx
+)
 
 if(NOT LLVM_TOOLCHAIN_CROSS_BUILD_MINGW)
     add_dependencies(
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -733,6 +761,27 @@ function(add_libcxx_libcxxabi_libunwind directory variant target_triple flags li
...
+    ExternalProject_Add_StepTargets(compiler_rt_${variant} check-compiler-rt)
+    ExternalProject_Add_StepDependencies(
+        compiler_rt_${variant}
+        check-compiler-rt
+        compiler_rt_${variant}-build
+    )
+    add_custom_target(check-compiler-rt-${variant})
+    add_dependencies(check-compiler-rt-${variant} compiler_rt_${variant}-check-compiler-rt)
+    add_dependencies(check-compiler-rt check-compiler-rt-${variant})
+    add_dependencies(check-llvm-toolchain-runtimes-${variant} check-compiler-rt-${variant})
```
mplatings, line 782: I think a similar `add_dependencies` would be appropriate near line 801 in the `add_libcxx_libcxxabi_libunwind_tests` function.

CMakeLists.txt
```diff
@@ -608,6 +623,14 @@ Hard-float library with target triple \"${target_triple}\" must end \"-eabihf\""
     endif()
 
     get_runtimes_flags("${directory}" "${flags}")
+    get_qemu_command("${target_triple}")
+
+    set(crt_test_emulator "${qemu_command} -L ${LLVM_BINARY_DIR}/${directory} ${qemu_params} -semihosting -nographic -kernel")
```
mplatings, line 628: I find the `crt_` prefix misleading so I suggest expanding to `compiler_rt_`

mplatings (changes requested):
Generally looks good, just a couple of tweaks please

mplatings (approved):

PR #241, opened 2023-05-11
Lit testing

pwprzybyla:
General concept of enabling lit testing for picolibc.

Do not forget to use external lit for example:
-DLLVM_DEFAULT_EXTERNAL_LIT=$PWD/llvm/bin/llvm-lit

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -456,7 +456,7 @@ function(add_picolibc directory variant target_triple flags qemu_params variant_
     configure_file(${CMAKE_CURRENT_FUNCTION_LIST_DIR}/cmake/meson-cross-build.txt.in ${BINARY_DIR}/meson-cross-build.txt @ONLY)
 
     # picolibc tests
-    string(REPLACE " " ";" qemu_params_list "${qemu_params}")
+    string(REPLACE " " ":" qemu_params_list "${qemu_params}")
     add_custom_target(check-picolibc_${variant})
     set(qemu_command "qemu-system-${cpu_family}")
     # Full list of picolibc tests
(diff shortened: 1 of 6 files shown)


mplatings:
> Do not forget to use external lit for example:
> -DLLVM_DEFAULT_EXTERNAL_LIT=$PWD/llvm/bin/llvm-lit

This is not acceptable.
Please change it so that it finds lit automatically.

pwprzybyla:
We have to use LLVM_DEFAULT_EXTERNAL_LIT explictly. 
Project by default use gcc to build llvm-toolchain, llvm internally is able to resolve path to lit and use it. But picolibc is external project and needs to be instructed where to find lit tool.

CMakeLists.txt
```diff
@@ -498,67 +498,65 @@ function(add_picolibc directory variant target_triple flags qemu_params variant_
...
-        list(REMOVE_ITEM picolibc_tests abort constructor-skip long_double math_errhandling malloc_stress test-except timegm)
+        list(REMOVE_ITEM picolibc_tests constructor-skip long_double math_errhandling malloc_stress timegm)
     endif()
     if(variant STREQUAL "armv8.1m.main_hard_nofp_mve")
-        list(REMOVE_ITEM picolibc_tests abort constructor-skip complex-funcs fenv ffs long_double math_errhandling math-funcs malloc_stress printf_scanf printf-tests rand regex test-efcvt test-except test-strtod timegm time-tests)
+        list(REMOVE_ITEM picolibc_tests constructor-skip complex-funcs fenv ffs long_double math_errhandling math-funcs malloc_stress printf_scanf printf-tests rand regex test-efcvt test-strtod timegm time-tests)
     endif()
 
+    configure_lit_site_cfg(${CMAKE_CURRENT_SOURCE_DIR}/test-support/picolibc-variant.cfg.in
+                            picolibc_tests_${variant}/lit.cfg.py @ONLY)
```
mplatings, line 538: Remove `@ONLY`. It's not used as an argument by configure_lit_site_cfg

CMakeLists.txt
```diff
@@ -498,67 +498,65 @@ function(add_picolibc directory variant target_triple flags qemu_params variant_
...
+        ${CMAKE_CURRENT_BINARY_DIR}/picolibc_tests_${variant}
+        DEPENDS picolibc_${variant} compiler_rt_${variant} llvm-toolchain-runtimes)
+
     function(picolibc_test test)
-        # -Oz parameter is added just like picolibc is doing, otherwise there is a build issue
-        # ld.lld: error: undefined symbol: __issignalingl  while building math_errhandling test.
-        add_custom_command(
-                OUTPUT picolibc_tests_${variant}/check-picolibc_${variant}_build_${test}.out
-                COMMAND ${LLVM_BINARY_DIR}/bin/clang${CMAKE_EXECUTABLE_SUFFIX} --config ${LLVM_BINARY_DIR}/bin/${variant}_semihost.cfg
+        set(BUILD_COMMAND_TMP ${LLVM_BINARY_DIR}/bin/clang${CMAKE_EXECUTABLE_SUFFIX} --config ${LLVM_BINARY_DIR}/bin/${variant}_semihost.cfg
```
mplatings, line 545: Instead of `${LLVM_BINARY_DIR}/bin/clang${CMAKE_EXECUTABLE_SUFFIX}` use `%clang`, similar to other lit tests.

You may need to update `picolibc-lit.site.cfg.in` to configure a `clang` variable.

mplatings, line 545: Instead of `--config ${LLVM_BINARY_DIR}/bin/${variant}_semihost.cfg` use `--config ${variant}_semihost.cfg`

This should work without further modifications.

CMakeLists.txt
```diff
@@ -498,67 +498,65 @@ function(add_picolibc directory variant target_triple flags qemu_params variant_
...
-        )
-        add_custom_target(
-                check-picolibc_${variant}-${test}
-                ${qemu_command} ${qemu_params_list} -semihosting -nographic -kernel check-picolibc_${variant}_build_${test}.out
-                DEPENDS picolibc_tests_${variant}/check-picolibc_${variant}_build_${test}.out
-                WORKING_DIRECTORY picolibc_tests_${variant}
-                COMMENT "TEST: ${variant} ${test}"
-        )
-        add_dependencies(check-picolibc_${variant} check-picolibc_${variant}-${test})
+                ${picolibc_SOURCE_DIR}/test/${test}.c ${ARGN} -o ${CMAKE_BINARY_DIR}/picolibc_tests_${variant}/check-picolibc_${variant}_build_${test}.out)
```
mplatings, line 548: Use `${CMAKE_CURRENT_BINARY_DIR}` instead of `${CMAKE_BINARY_DIR}`

CMakeLists.txt
mplatings, line 545: No need for separate BUILD_COMMAND_TMP and BUILD_COMMAND variables. You can use BUILD_COMMAND in both cases.

The same applies for TEST_COMMAND.

mplatings (changes requested):
I tried using these tests and they worked well. Just the code needs some tidying now please.

test-support/lit-exec-qemu.py
```diff
@@ -15,6 +15,27 @@
 # This script requires Python 3.6 or later
 assert sys.version_info >= (3, 6)
 
+picolibc_binary_name = "check-picolibc"
```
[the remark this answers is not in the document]
pwprzybyla: llvm tests have lit instructions in *.c and *.cpp files thus it is easy for them to use "not" tool for specific test. In this implementation we use template for all tests and it is more complicated. In my opinion one solution would be to keep a mapping in CMake file for each picolibc test name and explictly indicate to use "not". Or by doing this in python wrapper as i did.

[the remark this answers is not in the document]
mplatings: Another option is to use a different template for each test. It will be more repetitive, but if it makes things simpler then it could be worthwhile.

If you want to go the route of adding more logic in CMake then I suggest moving the picolibc test CMake code into a [subdirectory](https://cmake.org/cmake/help/latest/command/add_subdirectory.html) because the main CMakeLists.txt is quite bloated at this point.

mplatings (approved):
Looks good, thanks!

PR #242, opened 2023-05-19
Fix building from sources created by git archive

mplatings:
This fixes the source packages that GitHub automatically creates for each release.

Previously building from a source package would fail with:

`fatal: not a git repository (or any parent up to mount point /)`


voltur01 (approved):
LGTM, thank you for looking into this!

PR #245, opened 2023-06-06
tests: picolibc: math_errorhandling disable exception checks

domin144:
Disable checks for floating point exceptions.


mplatings:
If you think this option should be added in picolibc then please raise a pull request at https://github.com/picolibc/picolibc

PR #246, opened 2023-06-06
Remove LLVM_CCACHE_BUILD

amilendra:
PR #196 deprecated `LLVM_CCACHE_BUILD` and replaced it with `CMAKE_C_COMPILER_LAUNCHER` and `CMAKE_CXX_COMPILER_LAUNCHER`. However the CMake configuration still uses `LLVM_CCACHE_BUILD` which results in the CMake Warning:

```
Manually-specified variables were not used by the project:
  LLVM_CCACHE_BUILD
```

Remove use of `LLVM_CCACHE_BUILD` from the CMake configuration command.


amilendra:
> Looks correct as far as it goes, although to finish the job maybe it's work replacing these variables with `CMAKE_C_COMPILER_LAUNCHER` and `CMAKE_CXX_COMPILER_LAUNCHER`.
Thanks. I'll create another PR for that.

PR #249, opened 2023-06-09
Add bare-metal profiling sample

voltur01:
This sample is based on the discussion and the minimal runtime implementation provided by @smithp35 in https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/issues/197

### DIFF
--- a/samples/src/cpp-baremetal-semihosting-prof/README.md
+++ b/samples/src/cpp-baremetal-semihosting-prof/README.md
@@ -1,0 +1,15 @@
+# Code profiling and coverage sample
+
+This sample shows how to use instrumentation to emit profile data 
+and use it to show code coverage.
+
+This sample makes use of LLVM tools `llvm-profdata` and `llvm-cov` 
+that are not provided with LLVM Embedded Toolchain for Arm, 
+thus should be installed separately from the corresponding 
+version of the LLVM package https://releases.llvm.org/
+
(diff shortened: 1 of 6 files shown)


samples/src/cpp-baremetal-semihosting-prof/README.md
```diff
@@ -0,0 +1,15 @@
+# Code profiling and coverage sample
+
+This sample shows how to use instrumentation to emit profile data 
+and use it to show code coverage.
+
+This sample makes use of LLVM tools `llvm-profdata` and `llvm-cov` 
+that are not provided with LLVM Embedded Toolchain for Arm,
```
smithp35, line 7: Looking at https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/blob/main/CMakeLists.txt#L140 it looks like Linux and MacOS versions will, but I'm not sure about the mingw ones. Looks like something that we might be able to fix.

smithp35 (approved):
A couple of small suggestions surrounding the license and availability of some of the llvm tools. Otherwise LGTM.

samples/src/cpp-baremetal-semihosting-prof/README.md
voltur01: My bad, it is provided in all variants - let me fix this.

PR #250, opened 2023-06-12
Multilib

mplatings:
This change adds multilib configuration in accordance with the documentation in https://reviews.llvm.org/D143587

For the purposes of testing the multilib configuration, this change adds lit test configuration. It is intended that this test configuration will make it easier to add tests for other purposes as well, instead of adding targets in directly in CMake.


statham-arm (approved):

dcandler (approved):

mplatings:
Sorry, I had forgotten to pass `-Dmultilib=false` to the picolibc build system. (If that option is enabled then the picolibc build system tries to do its own multilib detection and havoc ensues).

statham-arm (approved):

PR #255, opened 2023-06-19
Run tests with llvm-lit

mplatings:
Previously tests were run as generic commands. When a test failed it would be hard to determine what exactly was failing without grepping the entire log. Using lit means test results can be output as JUnit XML which is understood natively by CI tools.

Another issue that this change addresses is that only the exit code of a test was checked. Now the output of the test is also checked.

Previously some tests were run on both build directory and the directory into which the toolchain has been installed. The benefit of this was questionable, so now those tests are only run on the latter.


CMakeLists.txt
```diff
@@ -880,7 +880,7 @@ function(add_library_variant target_arch)
     if(LLVM_TOOLCHAIN_LIBRARY_VARIANTS)
         if(NOT enable_${variant})
             message("Disabling library variant ${variant}")
-            continue()
+            return()
```
domin144, line 883: I missed this in last commit. Good you fixed this.

CMakeLists.txt
```diff
@@ -572,7 +572,7 @@ function(add_picolibc directory variant target_triple flags qemu_params variant_
 
     function(picolibc_test test)
         set(BUILD_PARAMS --config ${variant}_semihost.cfg
-                -T ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/tests/ldscripts/picolibc_${variant}.ld
+                -T ${CMAKE_CURRENT_FUNCTION_LIST_DIR}/test-support/ldscripts/picolibc_${variant}.ld
```
domin144, line 575: Just a note: the ld files will be removed with my change to run picolibc tests with meson.
If this change goes first, it can stay as is.

packagetest/cpp-baremetal-semihosting-ubsan.test
```diff
@@ -0,0 +1,5 @@
+# RUN: make -C %samples_dir/src/cpp-baremetal-semihosting-ubsan clean
+# RUN: make -C %samples_dir/src/cpp-baremetal-semihosting-ubsan run BIN_PATH=%unpack_directory/bin 2>&1 | FileCheck %s
+# RUN: make -C %samples_dir/src/cpp-baremetal-semihosting-ubsan clean
+# CHECK: UBSAN: add-overflow (recovered)
+# CHECK: C++ UBSan sample
```
domin144, line 5: missing line end

domin144 (commented):
The result looks cleaner.
I am glad that samples are still run with their makefiles. End user should not need to know how to use lit.
I tried to review as much as I could, but I am not an expert on lit tests.
BTW, ctest is also able to output junit xml output since version 3.21. Still, lit might more suited to compiler tests.

packagetest/cpp-baremetal-semihosting-ubsan.test
mplatings: Thanks, fixed.

PR #257, opened 2023-06-27
Extract get_test_executor function

domin144:
The code to construct the test executor was duplicated in three
functions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -435,12 +435,30 @@ add_custom_target(check-picolibc)
 # by add_subdirectory of llvm project
 set(LLVM_DEFAULT_EXTERNAL_LIT "${LLVM_BINARY_DIR}/bin/llvm-lit")
 
-function(get_qemu_command target_triple)
+function(get_test_executor target_triple qemu_machine qemu_cpu qemu_params)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -435,12 +435,31 @@ add_custom_target(check-picolibc)
...
+function(get_test_executor target_triple qemu_machine qemu_cpu qemu_params)
     if(target_triple MATCHES "^aarch64")
-        set(qemu_command "qemu-system-aarch64" PARENT_SCOPE)
+        set(qemu_command "qemu-system-aarch64")
     else()
-        set(qemu_command "qemu-system-arm" PARENT_SCOPE)
+        set(qemu_command "qemu-system-arm")
     endif()
+
+    # picolibc tests
```
mplatings, line 445: I don't understand why this comment is in this function

mplatings (commented):
It would be good to explain in the commit message why you're making the change

CMakeLists.txt
domin144: Good spot. I copied it over thoughtlessly. Done.

mplatings (approved):

PR #258, opened 2023-06-28
baremetal-uart sample: turn off QEMU semihosting.

statham-arm:
The whole point of this sample is that it runs without semihosting, so it's at least pointless to enable semihosting in the emulator, and perhaps it's also a testing hazard (in that if semihosting were accidentally included in the image, we'd prefer it to fail, so that we notice).


voltur01 (approved):
Good point, thank you!

PR #259, opened 2023-06-28
Multilib: add more implications between FPU types.

statham-arm:
Rather than doing things by half measures, I think I've now added _every_ subset/superset relation between the possible values of `-mfpu=thing`. This should futureproof multilib.yaml.in against further FPU-based confusion if more library variants are added, and also make it easier for downstreams to add extra library variants of their own.


mplatings (commented):
What would the code look like to generate these lists?
If it would involve lots of special cases then the hard-coded list might be preferable. But if it's just some set unions then I'd find that easier to understand.
Could be done in a separate Python script or in CMake using [list(REMOVE_DUPLICATES)](https://cmake.org/cmake/help/latest/command/list.html#remove-duplicates) to emulate set operations.

Also please add some tests. See test/multilib/armv8.1m.main.test as an example

statham-arm:
The steps to generate this data were roughly:
* extract the list of FPUs out of `llvm/include/llvm/TargetParser/ARMTargetParser.def` by looking for the first argument of each `ARM_FPU` definition
* exclude everything with `FPUVersion::NONE`, due to ABI incompatibility
* for each FPU, construct a clang driver command line using `-###`, and pull out the list of `-target-feature +this` and `-target-feature -that` from the `clang -cc1` subcommand
* for each pair of FPUs, decide whether the set of features in one is a superset of the other
* format the results of all that into YAML.

I did a lot of those steps manually or with editor macros, particularly the initial extraction of FPU names. The final double-loop over FPU pairs was done using a shell+Perl one-liner that I didn't keep (and if I had kept it I guarantee you wouldn't have enjoyed reading it :-).

I suppose we _could_ automate the whole thing, which would make it easier to re-run when the list of FPUs changes in future. But it's not work I've already done.

mplatings:
To satisfy my own curiosity, I wanted to understand what the relationships between FPUs, and the work that you've done here seems to be the best resource out there. I wrote some code to extract the relationships then encoded that in CMake:

```cmake
set(fpu_implications "")

function(fpu_implies fpu)
    foreach(implied ${ARGN})
        string(MAKE_C_IDENTIFIER ${implied} implied_variable)
        list(APPEND implied_fpus ${implied} ${${implied_variable}})
    endforeach()
    list(SORT implied_fpus)
    list(REMOVE_DUPLICATES implied_fpus)

    string(MAKE_C_IDENTIFIER ${fpu} implied_variable)
    set(${implied_variable} ${implied_fpus} PARENT_SCOPE)

    set(fpu_implications "${fpu_implications}
- Match: -mfpu=${fpu}
  Flags:"
    )

    foreach(implied ${implied_fpus})
        set(fpu_implications "${fpu_implications}
  - -mfpu=${implied}"
        )
    endforeach()

    set(fpu_implications "${fpu_implications}" PARENT_SCOPE)
endfunction()

fpu_implies(vfpv3xd-fp16
    vfpv3xd
)
fpu_implies(vfpv3-d16
    vfpv2
    vfpv3xd
)
fpu_implies(vfpv3-d16-fp16
    vfpv3-d16
    vfpv3xd-fp16
)
fpu_implies(vfpv3
    vfpv3-d16
)
fpu_implies(vfpv3-fp16
    vfpv3
    vfpv3-d16-fp16
)
fpu_implies(neon
    vfpv3
)
fpu_implies(neon-fp16
    vfpv3-fp16
    neon
)
fpu_implies(fpv4-sp-d16
    vfpv3xd-fp16
)
fpu_implies(vfpv4-d16
    vfpv3-d16-fp16
    fpv4-sp-d16
)
fpu_implies(vfpv4
    vfpv3-fp16
    vfpv4-d16
)
fpu_implies(neon-vfpv4
    vfpv4
    neon-fp16
)
fpu_implies(fpv5-sp-d16
    fpv4-sp-d16
)
fpu_implies(fpv5-d16
    vfpv4-d16
    fpv5-sp-d16
)
fpu_implies(fp-armv8-fullfp16-sp-d16
    fpv5-sp-d16
)
fpu_implies(fp-armv8-fullfp16-d16
    fpv5-d16
    fp-armv8-fullfp16-sp-d16
)
fpu_implies(fp-armv8
    vfpv4
    fpv5-d16
)
fpu_implies(neon-fp-armv8
    neon-vfpv4
    fp-armv8
)
fpu_implies(crypto-neon-fp-armv8
    neon-fp-armv8
)
```

The `fpu_implies` function isn't pretty but I found that the list of implications fit in my brain more easily. I've checked that `@fpu_implications@` would work as a drop-in substitute for the hard-coded list and is functionally equivalent. You're welcome to use it if you think appropriate. Either way please add tests.

mplatings:
I just saw https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/commit/c4f6ec363183111b2c181c86905d51e624a92100. Nice! Definitely tests needed :)

statham-arm:
I agree that your list looks nicer than my original hard-coded list! You've arranged to hand-write the transitive _reduction_ of the graph, whereas writing it directly into the output YAML I had to specify the transitive _closure_, including all the extra links that ought to be obvious.

But I prefer the autogenerated approach now that I've got it working, because it won't need hand-updating when a new FPU is added to LLVM. (Only if ARMTargetParser.def gets refactored or something similarly fundamental.)

In fact I wish we could do the same for architectures, although with all the `+foo` suffixes I think that would be considerably more difficult.

I'll add some tests to what I've got now, although I'm not sure how feasible it is to run them right now with the build broken...

mplatings (commented):
Terrifying, but aside from making the tests more concise LGTM

test/multilib/fpus.test
```diff
@@ -0,0 +1,26 @@
+# RUN: %clang -print-multi-flags-experimental --target=arm-none-eabihf -march=armv7m -mfpu=fpv5-d16 | FileCheck --check-prefix=VFPV4-D16 %s
+# RUN: %clang -print-multi-flags-experimental --target=arm-none-eabihf -march=armv7m -mfpu=fpv5-d16 | FileCheck --check-prefix=FPV4-SP-D16 %s
+# RUN: %clang -print-multi-flags-experimental --target=arm-none-eabihf -march=armv7m -mfpu=fpv5-d16 | FileCheck --check-prefix=FPV5-SP-D16 %s
```
mplatings, line 3:
```suggestion
# RUN: %clang -print-multi-flags-experimental --target=arm-none-eabihf -march=armv7m -mfpu=fpv5-d16 | FileCheck --check-prefixes=VFPV4-D16,FPV4-SP-D16,FPV5-SP-D16 %s
```

statham-arm: I tried this and it didn't work very well. The problem is that you have to arrange the checks in the right order: if you expect `-mfpu=fpv4-sp-d16` to appear in the output before `-mfpu=vfpv4-d16` then the two check lines have to be in that order. But the order of output from `-print-multi-flags[-experimental]` is not guaranteed, so that introduces fragility – if the output order changes, the test can break in a way that isn't actually interesting. With the test as I have it now, it's insensitive to the output ordering.

Separately from this, we discussed offline this morning rewriting the generation code in Python rather than cmake for legibility. I'm still planning to do that.

(Indeed, that could be a thing that changes the output order, because who knows what order the Python version will find it most convenient to generate!)

mplatings: I think you should assume that the order of `-print-multi-flags[-experimental]` is guaranteed. Many tests in llvm-project would break if it wasn't. See https://reviews.llvm.org/D142933#change-zuw0O32XbYXO

statham-arm: Ah, they're sorted by clang? I must admit I'd assumed that they would be emitted in whatever order `multilib.yaml` generated them, which in turn would depend on the generating script and its inputs. But fine, if clang is guaranteeing to sort them into a canonical order, then that's much easier.

CMakeLists.txt
```diff
@@ -1087,9 +1087,23 @@ add_library_variant(
 
 configure_file(
     ${CMAKE_CURRENT_SOURCE_DIR}/cmake/multilib.yaml.in
-    ${CMAKE_CURRENT_BINARY_DIR}/llvm/${TARGET_LIBRARIES_DIR}/multilib.yaml
+    ${CMAKE_CURRENT_BINARY_DIR}/multilib.yaml.in.manual
```
mplatings, line 1090: The filename could be made more obvious.
```suggestion
    ${CMAKE_CURRENT_BINARY_DIR}/multilib-without-fpus.yaml
```

mplatings (changes requested):

multilib-fpus.py
```diff
@@ -0,0 +1,226 @@
...
+    ]
+
+    raw_output = subprocess.check_output(command)
+
+    # The output of the above is a collection of JSON arrays each
+    # containing two strings, and each followed by a comma. Turn it
+    # into a single legal JSON declaration by appending a trailing
+    # null (so that there's something after the final comma) and
+    # wrapping it in array brackets.
+    json_output = b"[" + raw_output + b"null]"
```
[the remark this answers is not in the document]
statham-arm: (Bah, I had that version originally, and changed my mind about which was nicer)

multilib-fpus.py
```diff
@@ -0,0 +1,226 @@
...
+
+    # Now we can feed that to the JSON parser.
+    J = json.loads(json_output)
```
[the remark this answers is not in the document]
statham-arm: Actually, without the intervening removal of `None`, it's not really necessary to give it a name at all any more.

multilib-fpus.py
```diff
@@ -0,0 +1,226 @@
...
+    Strategy: the intended use of ARMTargetParser.def in the actual
+    LLVM build is to run it through the C preprocessor with whatever
+    #defines will generate the output you want. So the most reliable
+    way to get what _we_ want is to do exactly the same.
+
+    The output format I've chosen is actually JSON, because that's
+    close enough to C-like syntax that you can generate it easily
+    using cpp features like stringification, and also convenient for
+    Python to consume afterwards.
+
```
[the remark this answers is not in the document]
statham-arm: emacs puts that trailing blank line in whenever it rewraps a paragraph in a docstring. Never understood why!

CMakeLists.txt
statham-arm: Fine, although if we later add more autogeneration we'll need to rename it :-)

multilib-fpus.py
```diff
@@ -0,0 +1,214 @@
...
+        #
+        # The first argument will be a string literal giving the FPU
+        # name, which we copy into the JSON output still as a string
+        # literal.
+        #
+        # The third argument indicates the general FPU category, which
+        # we need so as to exclude FPUVersion::NONE. That is not
+        # already a string literal, so we stringify it in the
+        # preprocessor to make it legal JSON.
+        "-DARM_FPU(a,b,c,...)=[a,#c],",
```
mplatings, line 61: I made this comment before but GitHub seems to have lost it:
```suggestion
        "-DARM_FPU(name,kind,version,...)=[name,#version],",
```

mplatings (approved):

PR #254, opened 2023-06-15
picolibc: tests with meson

domin144:
Switch from manually building picolibc tests to bilding and running them
as part of picolibc build. This enables tests, which were not included
in the manual build.

Several fixes were necessary:

1. Add '-nostdlib' to picolibc c_args
This prevents errors like:
  ld.lld: error: unable to find library -lc
in checks for compiler flags support in picolibs build scripts.
Example of failing check:
  has_link_defsym = meson.get_cross_property('has_link_defsym', cc.has_link_argument('-Wl,--defsym=' + global_prefix + '_start=0'))

2. Make picolibc use compiler_rt in place of libgcc.

3. Change machines used with qemu and adjust memory layouts.
Some of picolibc tests need more memory, than previous machines
provided. Example error:
  ld.lld: error: section '.text' will not fit in region 'flash': overflowed by 34504 bytes

4. Postpone building picolibc tests to the test step. compiler_rt
requires picolibc to be installed. picolibc tests require compiler_rt.
The delayed tests enable picolibc to be installed before compiler_rt is built.

PR #261, opened 2023-06-29
Enable picolibc test constructor-skip

pwprzybyla:
Constructor skip test is another one that returns
non-zero value upon success


test-support/picolibc-lit-test-constructor-skip.template
```diff
@@ -0,0 +1,4 @@
+# Generated Picolibc lit for @variant@ @picolibc_SOURCE_DIR@/test/@test@.c
+# RUN: %clang @BUILD_PARAMS@
+# RUN: not @TEST_COMMAND@
+# CHECK-OUTPUT: errorcode=0
```
mplatings, line 4: This line doesn't do anything.
Is there any output from running the test successfully that you could check against?

mplatings:
Closing because #254 has landed.

PR #262, opened 2023-07-03
[picolibc] Remove fragment of patch

domin144:
The fragment is no longer needed as the issue it solved is now fixed upstream. Upstream fix: https://github.com/picolibc/picolibc/pull/499


mplatings (approved):
Thanks!

PR #274, opened 2023-07-17
Remove use of *.cfg files from samples.

voltur01:
Rely on the automatic multilib selection instead.

### DIFF
--- a/samples/Makefile.conf
+++ b/samples/Makefile.conf
@@ -19,7 +19,7 @@ ifndef BIN_PATH
 BIN_PATH=../../../bin
 endif
 
-MB_TARGET=--target=armv6m-none-eabi -march=armv6m -mfloat-abi=soft 
+MICROBIT_TARGET=--target=armv6m-none-eabi -march=armv6m -mfloat-abi=soft 
 AARCH64_TARGET=--target=aarch64-none-elf -march=armv8-a
 
 CRT=-lcrt0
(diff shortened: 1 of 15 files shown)


samples/Makefile.conf
```diff
@@ -18,3 +18,11 @@
 ifndef BIN_PATH
 BIN_PATH=../../../bin
 endif
+
+MB_TARGET=--target=armv6m-none-eabi -march=armv6m -mfloat-abi=soft
```
mplatings, line 22: I find the name `MB_TARGET` non-obvious. I suggest either `ARMV6M_TARGET` or `MULTIBIT_TARGET`

samples/Makefile.conf
```diff
@@ -18,3 +18,11 @@
 ifndef BIN_PATH
 BIN_PATH=../../../bin
 endif
+
+MB_TARGET=--target=armv6m-none-eabi -march=armv6m -mfloat-abi=soft 
+AARCH64_TARGET=--target=aarch64-none-elf -march=armv8-a
```
mplatings, line 23: `-march=armv8-a` could be removed. Up to you.

mplatings (changes requested):

samples/Makefile.conf
voltur01: Agreed, updated to MICROBIT_TARGET to be specific.

mplatings (approved):

PR #275, opened 2023-07-19
Remove generation of cfg files

voltur01:


mplatings (changes requested):
I think it would be nice to have config files to reduce the need for typing out `-fno-exceptions -fno-rtti -lcrt0` and likewise for semihosting.

The .cfg files are still referred to in README.md, packagetest & samples. Also please add this change to CHANGELOG.md.

voltur01:
Thank you for the review!

I hope that we will enable exception support soon, so having a special cfg file for disabling it may not be justified. For semihosting it might be still useful, however it would be similar amount of typing on the command line and require checking the docs to remember what to type - I would suggest to review usability as a separate task and as an option consider adding some specific cfg files.

I updated the packagetest files.

I updated README and CHANGELOG as well.

Samples were updated in https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/pull/274

mplatings (approved):
Ah, this change hasn't been rebase on top of #274 yet which is why I still saw the config files used in samples. LGTM.

PR #278, opened 2023-07-25
Fix compiler-rt tests

mplatings:
Since the .cfg files were removed, the compiler-rt tests have been failing. This change fixes the flags passed to those tests.


smithp35 (approved):
LGTM I can see that runtimes_flags sets the sysroot appropriately.

PR #283, opened 2023-08-01
On Mac create .dmg packages

mplatings:
This avoids the problem of macOS quarantining files that have been extracted from a .tar.gz.


voltur01 (approved):
LGTM

PR #284, opened 2023-08-02
XZ package compression

mplatings:
This reduces the size of the package file by about a third.
On a machine with many CPU cores, it reduces the elapsed time to perform the compression also by about a third.


voltur01 (approved):
LGTM, thank you for looking into this!

PR #287, opened 2023-08-02
add armv7-a and armv7-r targets

domin144:


CMakeLists.txt
```diff
@@ -966,7 +966,7 @@ add_library_variant(
 add_library_variant(
     armv6m
     SUFFIX soft_nofp
-    COMPILE_FLAGS "-mfloat-abi=soft -march=armv6m"
+    COMPILE_FLAGS "-mfloat-abi=soft -march=armv6-m"
```
mplatings, line 969: I'd prefer these changes weren't made. `armv6m` works with both `-march` and `--target` whereas `armv6-m` only works with the former.

CMakeLists.txt
```diff
@@ -977,10 +977,78 @@ add_library_variant(
     RAM_SIZE 2M
     STACK_SIZE 4K
 )
+add_library_variant(
+    armv7a
+    SUFFIX soft_nofp
+    COMPILE_FLAGS "-mfloat-abi=soft -march=armv7-a+nofp"
+    MULTILIB_FLAGS "--target=thumbv7-none-unknown-eabi -mfpu=none"
```
mplatings, line 984: I'm pretty sure this is incorrect. Please add tests in `test/multilib` to ensure these work and continue working.

mplatings (changes requested):

CMakeLists.txt
domin144: Ok. Reverting. The motivation was to be in line with GCC documentation.

CMakeLists.txt
domin144: You are right.
I added tests and fixed the flags. Thanks.

mplatings (approved):

PR #294, opened 2023-08-09
Use unstable libc++ ABI

mplatings:
This means users can benefit from all the latest libc++ improvements and bugfixes. For more information see
https://libcxx.llvm.org/DesignDocs/ABIVersioning.html

It is already the case that different versions of LLVM Embedded Toolchain for Arm are not expected to be ABI compatible.


smithp35 (commented):
I agree with the decision, although I think we should add a sentence to https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/blob/main/README.md#c-support 

Something like:
"LLVM Embedded Toolchain for Arm uses the unstable libc++ ABI version. This ABI uses all the latest libc++ improvements and bugfixes, but may result in link errors when linking against objects compiled against older versions of the ABI. For more information see  https://libcxx.llvm.org/DesignDocs/ABIVersioning.html."

mplatings:
@smithp35 good point, thanks. Done, and added an entry to the changelog as well.

amilendra (approved):

PR #297, opened 2023-08-17
Add LIBS_DEPEND_ON_TOOLS option

mplatings:
By default the build system will ensure tools like clang are up to date before using them to build libraries. If you are working on changes that are specific to libraries and unlikely to be affected by changes to tools then this behaviour is undesirable as you can spend a lot of time waiting for tools to be rebuilt. By setting LIBS_DEPEND_ON_TOOLS to false the dependency from libraries to tools is removed and the libraries can be built without rebuilding tools.


statham-arm (approved):
LGTM. Thanks – this will make my life easier!

Also, today I learned about `option` in CMake. I'd have done that more verbosely, as `set(thingy ON CACHE BOOL "description")`.

PR #308, opened 2023-09-20
Fix the profiling sample runtime

voltur01:
Recent upstream change causes __llvm_profile_raw_version to be defined by the compiler, thus no need to define it in the runtime.


smithp35 (approved):
Looks good to me. Thanks for the update. My guess at the upstream change is https://reviews.llvm.org/D157913

PR #309, opened 2023-09-25
Enable more Picolibc tests on aarch64

pwprzybyla:
Disable long double tests for aarch64
so more tests execute without failure.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -502,8 +502,10 @@ function(
 
     if(target_triple MATCHES "^aarch64")
         set(cpu_family aarch64)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -502,8 +502,10 @@ function(
 
     if(target_triple MATCHES "^aarch64")
         set(cpu_family aarch64)
+        set(enable_long_double_test false)
```
domin144, line 505: If the reason for failure of long double tests is known, it would be worth putting it in a commend here.

domin144 (approved):

PR #312, opened 2023-09-28
add ExclusiveGroup feature to multilib.yaml

### LINKED ISSUE #307: Simple C++ program cannot seem to find stdint.h for Cortex-M85
Using LLVM 17 Preview, I am trying to compile a simple program:

```
#include <stdint.h>

int main(void)
{
    uint8_t count = 0;
    uint32_t result = 0;

    for (count = 0; count < 10; count++)
    {
        result += count;
    }

    return result;
}
```

with command:

```
clang -std=c++11 -x c++ -mfloat-abi=hard -O2 -mcpu=cortex-m85 --target=arm-none-eabi -mthumb -c -o test.o test.cpp
```

If I change the above command to `mcpu=cortex-m33` or `mcpu=cortex-m4` it works fine - no errors. But with the above command, I get the following output:

```
clang version 17.0.0
Target: arm-none-unknown-eabi
Thread model: posix
InstalledDir: /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin
 (in-process)
 "/opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/clang-17" -cc1 -triple thumbv8.1m.main-none-unknown-eabihf -emit-obj -disable-free -clear-ast-before-backend -disable-llvm-verifier -discard-value-names -main-file-name test.cpp -mrelocation-model static -mframe-pointer=all -fmath-errno -ffp-contract=on -fno-rounding-math -mconstructor-aliases -nostdsysteminc -target-cpu cortex-m85 -target-feature -crc -target-feature -dotprod -target-feature +dsp -target-feature +mve -target-feature +mve.fp -target-feature +ras -target-feature -fp16fml -target-feature -bf16 -target-feature -sb -target-feature -i8mm -target-feature +lob -target-feature -cdecp0 -target-feature -cdecp1 -target-feature -cdecp2 -target-feature -cdecp3 -target-feature -cdecp4 -target-feature -cdecp5 -target-feature -cdecp6 -target-feature -cdecp7 -target-feature +pacbti -target-feature -hwdiv-arm -target-feature +hwdiv -target-feature +vfp2 -target-feature +vfp2sp -target-feature -vfp3 -target-feature +vfp3d16 -target-feature +vfp3d16sp -target-feature -vfp3sp -target-feature +fp16 -target-feature -vfp4 -target-feature +vfp4d16 -target-feature +vfp4d16sp -target-feature -vfp4sp -target-feature -fp-armv8 -target-feature +fp-armv8d16 -target-feature +fp-armv8d16sp -target-feature -fp-armv8sp -target-feature +fullfp16 -target-feature +fp64 -target-feature -d32 -target-feature -neon -target-feature -crypto -target-feature -sha2 -target-feature -aes -target-feature +strict-align -target-abi aapcs -mfloat-abi hard -Wunaligned-access -debugger-tuning=gdb -v -fcoverage-compilation-dir=/home/coder/workspace/peaks -resource-dir /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/lib/clang/17 -internal-isystem /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv8.1m.main_hard_fp/include/c++/v1 -internal-isystem /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv8m.main_hard_fp/include/c++/v1 -internal-isystem /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv7em_hard_fpv5_d16/include/c++/v1 -internal-isystem /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv7em_hard_fpv4_sp_d16/include/c++/v1 -internal-isystem /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/lib/clang/17/include -internal-isystem /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv8.1m.main_hard_fp/include -internal-isystem /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv8m.main_hard_fp/include -internal-isystem /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv7em_hard_fpv5_d16/include -internal-isystem /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv7em_hard_fpv4_sp_d16/include -O2 -std=c++11 -fdeprecated-macro -fdebug-compilation-dir=/home/coder/workspace/peaks -ferror-limit 19 -fno-signed-char -fgnuc-version=4.2.1 -fcxx-exceptions -fexceptions -fcolor-diagnostics -vectorize-loops -vectorize-slp -faddrsig -D__GCC_HAVE_DWARF2_CFI_ASM=1 -o test.o -x c++ test.cpp
clang -cc1 version 17.0.0 based upon LLVM 17.0.0-rc1 default target aarch64-linux-gnu
ignoring duplicate directory "/opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/lib/clang/17/include"
#include "..." search starts here:
#include <...> search starts here:
 /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv8.1m.main_hard_fp/include/c++/v1
 /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv8m.main_hard_fp/include/c++/v1
 /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv7em_hard_fpv5_d16/include/c++/v1
 /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv7em_hard_fpv4_sp_d16/include/c++/v1
 /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/lib/clang/17/include
 /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv8.1m.main_hard_fp/include
 /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv8m.main_hard_fp/include
 /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv7em_hard_fpv5_d16/include
 /opt/LLVMEmbeddedToolchainForArm-17.0.0-Linux-x86_64/bin/../lib/clang-runtimes/arm-none-eabi/armv7em_hard_fpv4_sp_d16/include
End of search list.
test.cpp:5:5: error: unknown type name 'uint8_t'
    5 |     uint8_t count = 0;
      |     ^
test.cpp:6:5: error: unknown type name 'uint32_t'
    6 |     uint32_t result = 0;
      |     ^
2 errors generated.
```

amilendra:
Fixes #307
This allows a multilib configuration to specify explicitly that a subset of its
libraries are alternatives to each other, and at most one of that subset should
be selected. To be consistent with existing semantics, the _last_ choice wins,
if more than one of the options matches the given flags.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -946,6 +946,7 @@ function(add_library_variant target_arch)
     foreach(flag ${multilib_flags_list})
         string(APPEND multilib_yaml_content "  - ${flag}\n")
     endforeach()
+    string(APPEND multilib_yaml_content "  ExclusiveGroup: stdlibs\n")
 
     install(
         DIRECTORY "${LLVM_BINARY_DIR}/${directory}/"
@@ -1085,21 +1086,6 @@ add_library_variant(
 )
 add_library_variant(
     armv7m
-    SUFFIX soft_nofp
-    COMPILE_FLAGS "-mfloat-abi=soft -march=armv7m -mfpu=none"
-    MULTILIB_FLAGS "--target=thumbv7m-none-unknown-eabi -mfpu=none"
-    QEMU_MACHINE "mps2-an385"
-    QEMU_CPU "cortex-m3"
-    BOOT_FLASH_ADDRESS 0x00000000
-    BOOT_FLASH_SIZE 0x1000
-    FLASH_ADDRESS 0x00001000
-    FLASH_SIZE 0x3ff000
-    RAM_ADDRESS 0x20000000
-    RAM_SIZE 2M
-    STACK_SIZE 4K
-)
-add_library_variant(
-    armv7m
     SUFFIX soft_fpv4_sp_d16
     COMPILE_FLAGS "-mfloat-abi=softfp -march=armv7m -mfpu=fpv4-sp-d16"
     MULTILIB_FLAGS "--target=thumbv7m-none-unknown-eabi -mfpu=fpv4-sp-d16"
@@ -1113,13 +1099,19 @@ add_library_variant(
     RAM_SIZE 2M
     STACK_SIZE 4K
 )
+# When no -mfpu=none is specified, the compiler internally adds all other
+# possible fpu settings before searching for matching variants. So for the
+# no-fpu variant to win, it has to be in multilab.yaml after all other
+# fpu variants. The order of variants in multilab.yaml depends on the order
+# of the add_library_variant calls. So the add_library_variant that adds
+# the soft_nofp for armv7m is placed after all other armv7m variants.
 add_library_variant(
-    armv7em
+    armv7m
     SUFFIX soft_nofp
-    COMPILE_FLAGS "-mfloat-abi=soft -march=armv7em -mfpu=none"
-    MULTILIB_FLAGS "--target=thumbv7em-none-unknown-eabi -mfpu=none"
-    QEMU_MACHINE "mps2-an386"
-    QEMU_CPU "cortex-m4"
+    COMPILE_FLAGS "-mfloat-abi=soft -march=armv7m -mfpu=none"
+    MULTILIB_FLAGS "--target=thumbv7m-none-unknown-eabi -mfpu=none"
+    QEMU_MACHINE "mps2-an385"
+    QEMU_CPU "cortex-m3"
     BOOT_FLASH_ADDRESS 0x00000000
     BOOT_FLASH_SIZE 0x1000
     FLASH_ADDRESS 0x00001000
@@ -1150,6 +1142,27 @@ add_library_variant(
     MULTILIB_FLAGS "--target=thumbv7em-none-unknown-eabihf -mfpu=fpv5-d16"
     QEMU_MACHINE "mps2-an500"
     QEMU_CPU "cortex-m7"
+    BOOT_FLASH_ADDRESS 0x00000000
+    BOOT_FLASH_SIZE 0x1000
+    FLASH_ADDRESS 0x00001000
+    FLASH_SIZE 0x3ff000
+    RAM_ADDRESS 0x20000000
+    RAM_SIZE 2M
+    STACK_SIZE 4K
+)
+# When no -mfpu=none is specified, the compiler internally adds all other
+# possible fpu settings before searching for matching variants. So for the
+# no-fpu variant to win, it has to be in multilab.yaml after all other
+# fpu variants. The order of variants in multilab.yaml depends on the order
+# of the add_library_variant calls. So the add_library_variant that adds
+# the soft_nofp for armv7em is placed after all other armv7em variants.
+add_library_variant(
+    armv7em
+    SUFFIX soft_nofp
+    COMPILE_FLAGS "-mfloat-abi=soft -march=armv7em -mfpu=none"
+    MULTILIB_FLAGS "--target=thumbv7em-none-unknown-eabi -mfpu=none"
+    QEMU_MACHINE "mps2-an386"
+    QEMU_CPU "cortex-m4"
     BOOT_FLASH_ADDRESS 0x00000000
     BOOT_FLASH_SIZE 0x1000
     FLASH_ADDRESS 0x00001000
--- a/patches/llvm-project.patch
+++ b/patches/llvm-project.patch
@@ -6,3 +6,259 @@ index 20c4f52cd378..eefdedbcb4f4 100644
 -#==============================================================================#
 +#==============================================================================#
  # This file specifies intentionally untracked files that git should ignore.
+diff --git a/clang/include/clang/Driver/Multilib.h b/clang/include/clang/Driver/Multilib.h
+index 1416559..6a9533e 100644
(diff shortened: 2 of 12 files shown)


CMakeLists.txt
```diff
@@ -1114,12 +1100,12 @@ add_library_variant(
     STACK_SIZE 4K
 )
 add_library_variant(
```
smithp35, line 1102: Looks like we've reordered
```
v7-m soft nofp
v7-m softfp fpv4-sp-d16
v7-em soft nofp
```
to
```
v7-m soft fpv4_sp_d16
v7-m soft nofp
v7-em soft nofp
```

This is explained by
```
When no -mfpu=none is specified, the compiler internally adds all other
possible fpu settings when searching for matching variants. So for the
no-fpu variant to win, it has to be added to multilab.yaml after all
other fpu variants. So reorder the add_library_variant calls accordingly.
```
I think it will be worth adding a comment to the CMake file, possibly above the first add_library_variant or here.

I'm also trying to work out if there is a problem with the selection code in the compiler. If there is, we would want to mark the comment with a FIXME: This is a workaround that can be removed when the selection logic is fixed.

The following is what I'd expect from the compiler, I don't think that we'll be able to fix in this patch.

Thinking about the selection logic there are 2 parts, not helped by the values of the mfloat-abi=softfp option which conflates the two:
* Can I support hardware floating point instructions? (controlled by -mfpu or inferred from -mcpu/-march)
* Can I pass floating point arguments in floating-point/SIMD extended registers? (ostensibly controlled by -mfloat-abi)

In an ideal world the two options would be independent of each other, but instead they are linked by a bunch of opaque implications.

I would expect that when there is no instance of -mfpu=none
* If the cpu/arch has no floating point unit or SIMD registers (integer only MVE) then the only possible answer is soft_nofp.
* If the cpu/arch has no floating point, but does have SIMD registers (integer only MVE) then there are two possible answers: soft_nofp and hard_nofp, with the latter selected by -mfloat-abi=hard.
* If the cpu/arch has floating point then there are 3 possible answers: soft_nofp with -mfloat-abi=soft, softfp with -mfloat-abi=softfp, and hard with -mfloat-abi=hard.

I would expect that with an instance of -mfpu=X then the compiler driver should add all variants that are "less than, i.e. compatible with X floating point units". The exception being:
* if -mfloat-abi=soft then do not pass through any floating point units (i.e. same as -mfpu=none)

I would expect that with no instance of -mfpu then the compiler driver should look up the default FPU from the CPU/Arch, and treat it in the same way as above.

We may need to look at what value of -mfloat-abi is being passed through. If there is no FPU then:
* softfp should never be passed, either an error or convert it to soft.
* hard is permitted if the CPU has SIMD registers (integer only MVE) otherwise error or conver to soft.

smithp35 (commented):
I can't see anything wrong, so this is OK to merge. I've made some small suggestions that could be delayed althouth I think it is important that we comment the changes to CMakeLists.txt. The others are just refactoring and could be delayed.

CMakeLists.txt
amilendra: Actually we reordered
```
armv7m soft nofp
armv7m soft fpv4_sp_d16
armv7em soft nofp
armv7em hard fpv4_sp_d16
armv7em hard fpv5_d16
```
to
```
armv7m soft fpv4_sp_d16
armv7m soft nofp
armv7em hard fpv4_sp_d16
armv7em hard fpv5_d16
armv7em soft nofp
```
I think the important thing is to have the nofp variant to be the last variant added for that particular CPU. So I'll add two comments before both of the `add_library_variant` calls that add the nofp variants.

amilendra:
I've added the comments to CMakeLists.txt. I'll leave the refactoring until we upstream this.

smithp35 (approved):
LGTM thanks for the update.

PR #313, opened 2023-09-28
Compiler's -S output and `-save-temps=obj` loses MVE feature which le…

simpal01:
At the moment clang is not generating the correct FPU directive for Arm V8.1m Mainline CPU's when assembled by -S. So when compiled with -S output and `-save-temps=obj` loses MVE feature which leads to assembler error.

There is a mismatch in the default FPU option that Arm V8.1-m CPUs define and what clang generates. For example, Clang generates fpv5-d16 for M85 and what is defined in ARMTargetParser.def is FK_FP_ARMV8_FULLFP16_D16.

This patch add cases for handling this FPU directive and thus makes clang generates the right FPU directives for Arm V8.1-m CPUs.

### DIFF
--- a/patches/llvm-project.patch
+++ b/patches/llvm-project.patch
@@ -262,3 +262,106 @@ index 2f86f8e..9c866db 100644
 +- Match: -mfloat-abi=softfp
 +  Flags: [-mfloat-abi=soft]
 +...
+diff --git a/clang/test/CodeGen/arm-v8.1m-check-fpu.ll b/clang/test/CodeGen/arm-v8.1m-check-fpu.ll
+new file mode 100644
+index 000000000000..5cff0457a190
+--- /dev/null
++++ b/clang/test/CodeGen/arm-v8.1m-check-fpu.ll
+@@ -0,0 +1,55 @@
++; REQUIRES: arm-registered-target
++; RUN: %clang --target=arm-none-eabi -mcpu=cortex-m85 -mfloat-abi=hard -save-temps=obj -S -o - %s | FileCheck %s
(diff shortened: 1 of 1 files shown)


smithp35 (approved):
LGTM, although I'm a bit close to the implementation. Would be good if we can some more eyes on it.

amilendra:
I'd rephrase this part of the commit message slightly.
> There is a change in the default FPU option that Arm V8.1-m CPUs defines and what the clang generates. For example, Clang generates fpv5-d16 for M85 and what defined in ARMTargetParser.def is FK_FP_ARMV8_FULLFP16_D16.

Please check if the below is what you meant.

There is a mismatch in the default FPU option that Arm V8.1-m CPUs define and what clang generates.
For example, Clang generates fpv5-d16 for M85 and what is defined in ARMTargetParser.def is FK_FP_ARMV8_FULLFP16_D16.

amilendra:
The implementation LGTM. I haven't checked if the test covers the newly added control-flow path. But if Simi has checked it LGTM.

simpal01:
> I'd rephrase this part of the commit message slightly.
> 
> > There is a change in the default FPU option that Arm V8.1-m CPUs defines and what the clang generates. For example, Clang generates fpv5-d16 for M85 and what defined in ARMTargetParser.def is FK_FP_ARMV8_FULLFP16_D16.
> 
> Please check if the below is what you meant.
> 
> There is a mismatch in the default FPU option that Arm V8.1-m CPUs define and what clang generates. For example, Clang generates fpv5-d16 for M85 and what is defined in ARMTargetParser.def is FK_FP_ARMV8_FULLFP16_D16.

Done

amilendra (approved):

PR #318, opened 2023-09-29
[main] Add trailing whitespace to patch without which the Windows build fails

amilendra:
Without the trailing whitespace the patch fails to apply on Windows.
Restore the trailing whitespace.


simpal01:
LGTM

PR #324, opened 2023-10-04
picolibc: update test wrapper to match upstream

domin144:
Two upstream changes:
c368f318 Remove c_args from per-multilib args
acc9b941 scripts: Change run-arm, run-aarch64 and run-riscv arg processing
changed the way comandline arguments are passed in semihosting tests.
Adjusted picolibc-test-wrappter.py to match run-arm used in upstream.


amilendra (approved):
LGTM Thanks.

PR #334, opened 2023-10-16
Add patches to make some libcxx armv7m tests pass

domin144:
Added patches from reviews in upstream.
https://reviews.llvm.org/D154246 (build)
https://reviews.llvm.org/D155521 (run with qemu)

Split llvm-project patches, so it is easier to track them.

Updated lit configuration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -244,7 +244,12 @@ endfunction()
 read_repo_version(llvmproject llvm-project)
 read_repo_version(picolibc picolibc)
 
-file(GLOB llvm_project_patches ${CMAKE_CURRENT_SOURCE_DIR}/patches/llvm-project/*.patch)
+set(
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -244,12 +244,13 @@ endfunction()
 read_repo_version(llvmproject llvm-project)
 read_repo_version(picolibc picolibc)
 
+file(GLOB llvm_project_patches ${CMAKE_CURRENT_SOURCE_DIR}/patches/llvm-project/*.patch)
```
pwprzybyla, line 247: According to CMake docs GLOB is not recommended. Please make sure it works in every use case.

domin144: I replaced the `GLOB` with explicit list.

pwprzybyla (approved):

PR #337, opened 2023-10-17
Enable exceptions in libcxxabi and libcxx.

pwprzybyla:


CMakeLists.txt
```diff
@@ -1144,6 +1170,8 @@ function(add_library_variant target_arch)
             "${lit_test_executor}"
             "${LLVM_TOOLCHAIN_C_LIBRARY}_${variant}-install"
         )
+        set(enable_exceptions OFF)
```
statham-arm, line 1173: It's a neat trick to use dynamic scope to pass these parameters into `add_libcxx_libcxxabi_libunwind` without having to include them in the function's parameter list, but I'm not sure it improves readability. I think it would be better to make them function parameters in the usual way, so that callers _know_ they have to think about which way to set them.

This means that in our downstream code that also calls that function, we'll have to make a corresponding change, but I think that's the best thing in the long run.

CMakeLists.txt
```diff
@@ -1153,13 +1181,47 @@ function(add_library_variant target_arch)
             "${LLVM_TOOLCHAIN_C_LIBRARY}_${variant}-install"
             "${${LLVM_TOOLCHAIN_C_LIBRARY}_specific_runtimes_options}"
         )
+        if(LLVM_TOOLCHAIN_C_LIBRARY STREQUAL picolibc)
```
statham-arm, line 1184: This is quite a lot of repetitive code, and makes two more places that we'll have to modify any further changes to this function call. Can we turn it into a loop? I know it's annoying to loop over pairs of variables in cmake, but something like this should work, popping two elements at a time from an interleaved list:
```
set(exc_rtti_settings OFF OFF)
if(LLVM_TOOLCHAIN_C_LIBRARY STREQUAL picolibc)
    set(exc_rtti_settings ${exc_rtti_settings} ON OFF ON ON)
endif()
while(exc_rtti_settings)
    list(POP_FRONT exc_rtti_settings enable_exceptions enable_rtti)
    add_libcxx_libcxxabi_libunwind(...)
endwhile()
```

Better still, we could set up `exc_rtti_settings` once, and then use it for this loop _and_ the loop below that adds tests (making a copy so that POP_FRONT doesn't destroy it the first time).

statham-arm (changes requested):

PR #339, opened 2023-10-20
Enable cxx exceptions and rtti.

pwprzybyla:
Exceptions for libcxx and rtti has to be enabled together.

### DIFF
--- a/patches/llvm-project/0004-xfail-two-remaining-libcxx-with-picolibc-tests.patch
+++ b/patches/llvm-project/0004-xfail-two-remaining-libcxx-with-picolibc-tests.patch
@@ -38,6 +38,34 @@ index 5a70ea5bd570..92dfbe8eb1ab 100644
  #include <cstdlib>
  
  void f() {}
+diff --git a/libcxx/test/std/re/re.alg/re.alg.match/exponential.pass.cpp b/libcxx/test/std/re/re.alg/re.alg.match/exponential.pass.cpp
+index 2493f6a643a6..dae7d8640622 100644
+--- a/libcxx/test/std/re/re.alg/re.alg.match/exponential.pass.cpp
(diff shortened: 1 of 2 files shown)


statham-arm (approved):

patches/llvm-project/0004-xfail-two-remaining-libcxx-with-picolibc-tests.patch
```diff
@@ -29,15 +29,43 @@ index 5a70ea5bd570..92dfbe8eb1ab 100644
 --- a/libcxx/test/std/language.support/support.start.term/quick_exit.pass.cpp
 +++ b/libcxx/test/std/language.support/support.start.term/quick_exit.pass.cpp
 @@ -16,6 +16,9 @@
- 
+
```
statham-arm, line 32: I'm suspicious that this patch file seems to have had trailing whitespace stripped. Are you sure it still applies correctly like this? Normally I expect that the leading space on a patch line (indicating "neither added nor removed") is mandatory even when the rest of the line is empty. A _completely_ empty line in a patch file is a syntax error.

Same goes for the `-- ` sig-separator at the end of the patch file, added by `git format-patch`.

pwprzybyla: I have edited this patch in an IDE. I belive it changed the whitespaces. However on linux patch applies without issues.

pwprzybyla: I can confirm that trailing spaces were removed by IDE upon save of a file. I have enabled file.TrimTrailingWhitespace option.

statham-arm (approved):
Looks much better, thanks!

PR #354, opened 2023-11-13
Fix test failure on CI

pwprzybyla:
Refactor libraries with exceptions
    
    Rework of libraries with exceptions
    so one can use them with --sysroot parameter
    to clang compiler.


amilendra:
I don't think the failures are CI specific (you probably can reproduce them in a non-CI environment as well) so having the word CI in the commit message and subject has much utility. So replace it with details of the tests that failed (i.e. package tests)

amilendra (approved):

PR #362, opened 2023-11-23
Need Changes in CMakeLists.txt to match with the new patch filename.

simpal01:


CMakeLists.txt
```diff
@@ -256,7 +256,7 @@ read_repo_version(${LLVM_TOOLCHAIN_C_LIBRARY} ${LLVM_TOOLCHAIN_C_LIBRARY})
 
 set(
     llvm_project_patches
-    ${CMAKE_CURRENT_SOURCE_DIR}/patches/llvm-project/0001-Fix-llvm-test-CodeGen-ARM-build-attributes.ll-test.patch
+    ${CMAKE_CURRENT_SOURCE_DIR}/patches/llvm-project/0001-Add-ExclusiveGroup-feature-to-multilib.yam.patch
```
pwprzybyla, line 259: I think it should be named yaml.patch not yam.patch?

domin144: I think the name is generated by git based on the commit message. There is probably some length limit git imposes. I would prefer to keep the generated name, so we don't need to rename again if the patch will be regenerated.

pwprzybyla (approved):

voltur01 (approved):

PR #367, opened 2023-11-29
Fix multilib selection process

pwprzybyla:
Fix multilib selection process
so process will respect
-fno-exception and -fno-rtti multilib flags

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1138,6 +1138,10 @@ function(add_library_variant target_arch)
         set(variant "${target_arch}")
     endif()
 
+    if(NOT VARIANT_ENABLE_RTTI_EXCEPTIONS)
+        set(VARIANT_MULTILIB_FLAGS "${VARIANT_MULTILIB_FLAGS} -fno-exceptions -fno-rtti")
+    endif()
+
     if(LLVM_TOOLCHAIN_LIBRARY_VARIANTS)
         if(NOT enable_${variant})
             message("Disabling library variant ${variant}")
@@ -1237,21 +1241,6 @@ set(multilib_yaml_content "")
 # the real flash.
 add_library_variant(
     aarch64
-    COMPILE_FLAGS "-march=armv8-a"
-    MULTILIB_FLAGS "--target=aarch64-none-unknown-elf"
-    QEMU_MACHINE "virt"
-    QEMU_CPU "cortex-a57"
-    BOOT_FLASH_ADDRESS 0x40000000
-    BOOT_FLASH_SIZE 0x1000
-    FLASH_ADDRESS 0x40001000
-    FLASH_SIZE 0xfff000
-    RAM_ADDRESS 0x41000000
-    RAM_SIZE 0x1000000
-    STACK_SIZE 8K
-    ENABLE_RTTI_EXCEPTIONS OFF
-)
-add_library_variant(
-    aarch64
     SUFFIX exceptions_rtti
     COMPILE_FLAGS "-march=armv8-a"
     MULTILIB_FLAGS "--target=aarch64-none-unknown-elf"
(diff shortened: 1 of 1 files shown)


patches/llvm-project/0015-Multilib-support-for-libraries-with-exceptions.patch
```diff
@@ -0,0 +1,27 @@
...
+diff --git a/clang/lib/Driver/ToolChains/BareMetal.cpp b/clang/lib/Driver/ToolChains/BareMetal.cpp
+index 42c8336e626c..d887c650748d 100644
+--- a/clang/lib/Driver/ToolChains/BareMetal.cpp
++++ b/clang/lib/Driver/ToolChains/BareMetal.cpp
+@@ -179,6 +179,10 @@ static void findMultilibsFromYAML(const ToolChain &TC, const Driver &D,
+   if (!MB)
+     return;
+   Multilib::flags_list Flags = TC.getMultilibFlags(Args);
++  if (TC.getRTTIMode() == ToolChain::RTTIMode::RM_Disabled) {
++    Flags.push_back("-fno-exceptions");
```
voltur01, line 19: I believe exceptions and RTTI can be enabled/disabled independently - we can choose to combine as a library variant, but we should not hardcode them here, I think.

patches/llvm-project/0015-Multilib-support-for-libraries-with-exceptions.patch
```diff
@@ -0,0 +1,27 @@
+From 26fe33c4db8b786077e34966a4d9e8b40dc90745 Mon Sep 17 00:00:00 2001
+From: Piotr Przybyla <piotr.przybyla@arm.com>
+Date: Wed, 29 Nov 2023 14:05:00 +0000
+Subject: [PATCH 15/15] Multilib support for libraries with exceptions
+
+---
+ clang/lib/Driver/ToolChains/BareMetal.cpp | 4 ++++
```
voltur01, line 7: Should we fix in this upstream, indead of adding a patch?

patches/llvm-project/0015-Multilib-support-for-libraries-with-exceptions.patch
pwprzybyla: When exceptions are enabled without rtti one of the tests fails. This is the reason we have it always enabled/disabled at the same time. More to say we ship only two types of library. Possibility to enable/disable rtti and exceptions independently would require to ship more library variants.

patches/llvm-project/0015-Multilib-support-for-libraries-with-exceptions.patch
pwprzybyla: Yes we can do it.

patches/llvm-project/0015-Multilib-support-for-libraries-with-exceptions.patch
domin144: One thing is what we do and the other is what other users of the upstream code will do. Even if there are dependencies between the two elsewhere, I agree the checks should be separate here.

Could we also have explicit flags for the positive case? I mean `-fexceptions` and `-frtti`?

As the current version works for us, I think it can be fixed during upstream review.

CMakeLists.txt
```diff
@@ -1252,13 +1253,12 @@ add_library_variant(
...
     STACK_SIZE 8K
-    ENABLE_RTTI_EXCEPTIONS OFF
+    ENABLE_RTTI_EXCEPTIONS ON
 )
 add_library_variant(
     aarch64
-    SUFFIX exceptions_rtti
     COMPILE_FLAGS "-march=armv8-a"
-    MULTILIB_FLAGS "--target=aarch64-none-unknown-elf"
+    MULTILIB_FLAGS "--target=aarch64-none-unknown-elf -fno-exceptions -fno-rtti"
```
domin144, line 1261: Wouldn't it be better to add this inside `add_library_variant` based on the `ENABLE_RTTI_EXCEPTIONS` setting?
BTW: Maybe it would be useful to also add `-fexceptions` and `-frtti` options, when the extensions are enabled.

CMakeLists.txt
```diff
@@ -1241,6 +1241,7 @@ set(multilib_yaml_content "")
 # the real flash.
 add_library_variant(
     aarch64
+    SUFFIX exceptions_rtti
```
domin144, line 1244: If the order of variants with and without exceptions is important, it would be nice to have a comment on that.

patches/llvm-project/0015-Multilib-support-for-libraries-with-exceptions.patch
voltur01: We will have one set of libraries for now, but this should be addressed in the multilib mapping file - the driver should allow splitting them later, i.e. should allow configuring exceptions and rtti flags independently.

domin144: As for the test. I think the test should be fixed upstream with something like:
```
// REQUIRES: rtti
```

CMakeLists.txt
domin144: Thanks, I missed it.

PR #370, opened 2023-12-01
Update for upstreaming of multilib exclusive groups.

statham-arm:
The ExclusiveGroup feature we added to multilib.yaml in our previous patch 0001 has now been upstreamed, with slightly different syntax in multilib.yaml. So we must remove that patch, and also adjust our multilib.yaml setup to match the changed syntax.

The new llvm-project patch series applies cleanly against upstream commit da86d4a8c956f0fcee21444eb6de9f05d39d6574.


smithp35 (approved):
LGTM

PR #380, opened 2024-01-19
Disable small versions of ctype functions

amilendra:
https://github.com/picolibc/picolibc/commit/43a3894158561b568118d31376edb465238fd320 adds support for a small ctype implementation -- an inline version which performs direct character value comparisons without needing a table.

However this implementation does not support locales or extended character sets. Therefore fallback to the classic, table based ctype implementation.

### DIFF
--- a/patches/picolibc.patch
+++ b/patches/picolibc.patch
@@ -1,8 +1,8 @@
 diff --git a/meson.build b/meson.build
-index c49856ee0..3e927ea75 100644
+index d82736966..55105b25c 100644
 --- a/meson.build
 +++ b/meson.build
-@@ -972,6 +972,12 @@ if get_option('newlib-retargetable-locking') != get_option('newlib-multithread')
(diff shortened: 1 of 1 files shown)


patches/picolibc.patch
```diff
@@ -11,6 +11,13 @@ index c49856ee0..3e927ea75 100644
 +It's necessary to set this globally because inline functions in
 +libc++ headers call the GNU functions.'''
 +)
++
++conf_data.set('_PICOLIBC_CTYPE_SMALL', '0',
++        description: '''Disable small versions of ctype functions.
++Use this to use the classic, table based ctype implementation and not
```
statham-arm, line 17: This text seems to be a description of what `_PICOLIBC_CTYPE_SMALL` is for. Wouldn't it be more useful to describe _why_ we want to set it in a non-default way? The purpose of the flag can be found out from picolibc itself.

Something like: "libc++ expects newlib-style ctype tables, and also expects support for locales and extended character sets, so picolibc's small ctype is not compatible with it".

statham-arm (approved):

PR #385, opened 2024-02-14
More tests for multilib selection

domin144:
Added failing test for softfp + armv8.1m.main+mve.
Added {{$}} not to match longer variant names.


domin144:
Just the second commit. The first one is reviewed in separate PR #384 .

pwprzybyla (approved):

PR #390, opened 2024-02-22
picolibc: linker script: Add '__lcxx_override' section

domin144:
This section was added into libc++ and conains the "new" operator[1]. Placing the "new" operator in the section allows checking, if the operator (defined as weak symbol) was overridden.

The check relies on the "__start_" and "__end_" symbols generated by linker. These symbols would not be generated for names, which are not valid C identifiers. For this reason name like ".text.__lcxx_generated" cannot be used.

[1] https://github.com/llvm/llvm-project/pull/69498

### DIFF
--- a/patches/picolibc.patch
+++ b/patches/picolibc.patch
@@ -1,8 +1,8 @@
 diff --git a/meson.build b/meson.build
-index d82736966..55105b25c 100644
+index 0fdfa0412..8e679f166 100644
 --- a/meson.build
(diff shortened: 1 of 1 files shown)


statham-arm (approved):
LGTM, although I wonder if it might be a good idea to pre-emptively match `__lcxx_*`? If libc++ upstream likes that naming scheme, they might use it again for the next special case they find.

PR #391, opened 2024-02-23
picolibc: linker script: force start/stop symbols for __lcxx_override

domin144:
The __start___lcxx_override and __stop___lcxx_override symbols were not generated for some tests. This change manually defines those symbols to fix following link error:
```
  ld.lld: error: undefined symbol: __start___lcxx_override
  >>> referenced by stdlib_new_delete.cpp
```

### DIFF
--- a/patches/picolibc.patch
+++ b/patches/picolibc.patch
@@ -22,14 +22,16 @@ index 0fdfa0412..8e679f166 100644
  	      cc.has_argument('-fno-tree-loop-distribute-patterns'),
  	      description: 'Compiler flag to prevent detecting memcpy/memset patterns')
 diff --git a/picolibc.ld.in b/picolibc.ld.in
-index b97ea3300..71cf51afe 100644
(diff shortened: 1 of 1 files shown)


domin144:
The previous "fix" broke some other tests. The error is as stated above.

I checked one of the failing tests (make_error_code.pass.cpp) and it does not do anything special. Especially, it does not look like doing any calls to new operator.

My theory is that linker optimizes out all the functions in the `_lcxx_override` section and thus does not generate the `__start___lcxx_override` and `__stop___lcxx_override` symbols. In such case, I would expect the functions using the two symbols to also be optimized, as they are only called from inside the new/delete operators. Somehow, linker still wants the two symbols.

I empirically made (and checked with full 2 hour test run this time) this PR for now, but I hope more detailed analysis will follow, when trying to find solution in the upstream.

pwprzybyla (approved):

PR #392, opened 2024-02-27
newlib: Enable _DEFAULT_SOURCE globally

domin144:
libc++ std::print function implementation relies on access to POSIX function "fileno", called in __is_posix_terminal. newlib does not expose that function in stdio.h header, unless __POSIX_VISIBLE is defined. Adding _DEFAULT_SOURCE macro enables the __POSIX_VISIBLE macro.

The flag was not added to LIBCXX_EXTRA_SITE_DEFINES, as the stdio.h header could be included before libc++ headers. In such case stdio.h would be parsed without the __POSIX_VISIBLE define and would not be parsed again, because of it's header guards.

I removed forced declaration of `FILE` in `wchar.h` as `_DEFAULT_SOURCE` already enables that.

### DIFF
--- a/patches/newlib.patch
+++ b/patches/newlib.patch
@@ -1,5 +1,5 @@
 diff --git a/libgloss/arm/cpu-init/rdimon-aem.S b/libgloss/arm/cpu-init/rdimon-aem.S
-index 95b86e4d..b91034ae 100644
+index 95b86e4d4..b91034ae6 100644
 --- a/libgloss/arm/cpu-init/rdimon-aem.S
 +++ b/libgloss/arm/cpu-init/rdimon-aem.S
 @@ -60,7 +60,7 @@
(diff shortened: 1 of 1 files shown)


patches/newlib.patch
```diff
@@ -360,21 +360,21 @@ index 7c23c7a0..2fc58416 100755
...
+ #include <_newlib_version.h>
  
--#if __POSIX_VISIBLE >= 200809 || _XSI_VISIBLE
-+#if 1 // __POSIX_VISIBLE >= 200809 || _XSI_VISIBLE
- /* As in stdio.h, <sys/reent.h> defines __FILE. */
- #if !defined(__FILE_defined)
- typedef __FILE FILE;
++#define _DEFAULT_SOURCE
++
+ /* Macro to test version of GCC.  Returns 0 for non-GCC or too old GCC. */
```
eleanor-arm, line 373: This part seems unrelated to the purpose of the patch

eleanor-arm: I think I see what you mean. In the "File Changed" view, it looks like the GCC version test has been added but not if I open the `newlib.patc` file completely  👍

PR #395, opened 2024-02-28
Include runtime check targets

pwprzybyla:
Include runtime check targets into check-llvm-toolchain


CMakeLists.txt
```diff
@@ -1210,6 +1212,7 @@ function(add_library_variant target_arch)
             message("C++ runtime libraries tests disabled for ${variant}")
         else()
             add_custom_target(check-llvm-toolchain-runtimes-${variant})
+            add_dependencies(check-llvm-toolchain check-llvm-toolchain-runtimes-${variant})
```
domin144, line 1215: Wouldn't it be simpler to just make `check-llvm-toolchain` depend on `check-llvm-toolchain-runtimes`, than to rely on each variant individually?

domin144 (approved):

PR #398, opened 2024-03-06
Separate the flags for exceptions and rtti.

domin144:
Updated the documentation as exceptions and rtti are supported now.

Added add_library_variants_for_cpu function to reduce duplication.


voltur01 (approved):
LGTM, thanks!

PR #405, opened 2024-03-22
Add verbose and debug options to qemu wrappers

domin144:


test-support/lit-exec-qemu.py
```diff
@@ -72,6 +83,8 @@ def main():
         [args.image] + args.arguments,
         args.timeout,
         args.execdir,
+        args.verbose,
+        args.gdb_servser,
```
statham-arm, line 87: Surely `args.gdb_server`?

test-support/lit-exec-qemu.py
```diff
@@ -55,6 +55,17 @@ def main():
...
         help="ignored, used for compatibility with libc++ tests",
     )
+    parser.add_argument(
+        "--verbose",
+        action="store_true",
+        help="Print verbose output. This may affect test result, as the output "
+        "will be added to the output of the test.",
+    )
+    parser.add_argument(
+        "--gdb-server",
```
statham-arm, line 65: Could we make this option able to specify a gdb server location other than a port on localhost? The standard TCP-based gdbserver is a security hole, and always has been (it allows another user on the same system to connect to your gdbserver).

Setting up gdb another way is awkward, but not impossible. If a user is cautious enough to go to the effort, I'd rather this script didn't put an extra obstacle in their way.

From the `qemu-system-arm` man page, I think it would be enough to allow the user to specify their own choice of argument to qemu's `-gdb` option.

(That way, you can specify a pty instead of a TCP port. My current best solution to secure intra-system gdbserver involves two ptys joined by a user process copying data back and forth.)

statham-arm (changes requested):

test-support/lit-exec-qemu.py
domin144: Yes, thanks!

test-support/lit-exec-qemu.py
domin144: There already is an option to pass any option to qemu, e.g. "--qemu-params=-S:-s". I think it will be better to remove the "--gdb-server" option completely and let user use the qemu switches - they are more flexible and better documented.
I will leave the verbose flag, though.

Issue #404, opened 2024-03-21
[CMakeLists] git command does not work on linux


sivan-shani:
CMakeLists.txt instructs how to build by checking out and patch the repos manually.
In the instructions, the command:
`git -C repos/llvm-project am -k ../../patches/llvm-project/*.patch`
Does not work on Linux (probably because the the shell expands the '*' and not git)

voltur01:
Hi Sivan,

You are right, looks like the pattern is expanded and used in different directories, thus the relative path is wrong in one of them. Providing the full path woks for me, e.g.

git -C repos/llvm-project am -k `pwd`/../patches/llvm-project/*.patch

I will update the doc, thank you for reporting this!

PR #408, opened 2024-03-26
Fix example build from source command lines

voltur01:
Fixes https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/issues/404

### DIFF
--- a/docs/building-from-source.md
+++ b/docs/building-from-source.md
@@ -61,9 +61,9 @@ If you check out repos manually then it is your responsibility to ensure that th
 ```
 mkdir repos
 git -C repos clone https://github.com/llvm/llvm-project.git
-git -C repos/llvm-project am -k ../../patches/llvm-project/*.patch
+git -C repos/llvm-project am -k `pwd`/../patches/llvm-project/*.patch
 git -C repos clone https://github.com/picolibc/picolibc.git
-git -C repos/picolibc apply ../../patches/picolibc.patch
+git -C repos/picolibc apply `pwd`/../patches/picolibc.patch
 mkdir build
 cd build
 cmake .. -GNinja -DFETCHCONTENT_SOURCE_DIR_LLVMPROJECT=../repos/llvm-project -DFETCHCONTENT_SOURCE_DIR_PICOLIBC=../repos/picolibc
--- a/docs/building-from-source.md
+++ b/docs/building-from-source.md
@@ -61,9 +61,9 @@ If you check out repos manually then it is your responsibility to ensure that th
 ```
 mkdir repos
 git -C repos clone https://github.com/llvm/llvm-project.git
-git -C repos/llvm-project am -k `pwd`/../patches/llvm-project/*.patch
+git -C repos/llvm-project am -k "$(PWD)"/../patches/llvm-project/*.patch
 git -C repos clone https://github.com/picolibc/picolibc.git
-git -C repos/picolibc apply `pwd`/../patches/picolibc.patch
+git -C repos/picolibc apply "$(PWD)"/../patches/picolibc.patch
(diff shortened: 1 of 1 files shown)


docs/building-from-source.md
```diff
@@ -61,9 +61,9 @@ If you check out repos manually then it is your responsibility to ensure that th
 ```
 mkdir repos
 git -C repos clone https://github.com/llvm/llvm-project.git
-git -C repos/llvm-project am -k ../../patches/llvm-project/*.patch
+git -C repos/llvm-project am -k `pwd`/../patches/llvm-project/*.patch
```
statham-arm, line 64: The backticked `pwd` invocation ought to be in double quotes, to protect against special characters like spaces in the output of the subprocess.

Also it may be possible to avoid the subprocess invocation completely, by using `"$PWD"`.

(If you do have to use a subprocess, my preference is to use `$(...)` rather than backticks, but the main reason is that they nest more sensibly, which isn't particularly important here. Even so, if it were me, I'd try to model good habits at every opportunity.)

statham-arm (changes requested):

AlfieRichardsArm:
Why does this need to construct a absolute path rather than using a relative path?

ie. using
```
`pwd`/../patches/llvm-project/*.patch
```
Instead of the more standard
```
../patches/llvm-project/*.patch
```
which should do the same thing?

statham-arm:
That part makes sense to me. It's because the overall command is using `git -C`, meaning "change to some other directory before running the git command". So the relative path `../foo` would be interpreted relative to the cwd _after_ git changes dir. But we want a path relative to where we are _before_ git changes dir.

AlfieRichardsArm:
Ahhh thank you @statham-arm, that makes sense

docs/building-from-source.md
voltur01: Great points! Fixed.

docs/building-from-source.md
```diff
@@ -61,9 +61,9 @@ If you check out repos manually then it is your responsibility to ensure that th
 ```
 mkdir repos
 git -C repos clone https://github.com/llvm/llvm-project.git
-git -C repos/llvm-project am -k ../../patches/llvm-project/*.patch
+git -C repos/llvm-project am -k "$(PWD)"/../patches/llvm-project/*.patch
```
statham-arm, line 64: Sorry, but you've combined parts of both my suggestions into something that doesn't work. (Did you re-test it, even as a standalone command?)

`"$PWD"` means: expand the shell variable `PWD`, which the shell should already be keeping up to date with your current directory. `"${PWD}"` would also work: the braces explicitly delimit the variable name in case of ambiguity, although in this case there's no ambiguity in the first place.

`"$(pwd)"` means: run the subcommand `pwd`, and substitute its output.

But `"$(PWD)"` means run the subcommand `PWD`, and _that_ subcommand doesn't exist, because Unix is case sensitive:
```
$ echo "$(PWD)"
PWD: command not found
```

statham-arm (changes requested):

docs/building-from-source.md
voltur01: Sorry, this is a case of works-on-my-pc, because apparently I have:
$ which PWD
/bin/PWD

Let me fix it again.

statham-arm: Ah, I bet that was on a Mac? MacOS is like Unix except case insensitive (by default), so yes, if you test on a Mac, you can miss all sorts of case-related problems that show up everywhere else.

voltur01: Yes, I am on macOS.

statham-arm (approved):

PR #422, opened 2024-04-19
Port proflib.c to LLVM-18 INSTR_PROF_RAW_VERSION 9

smithp35:
There have been a number of changes to the profile output since LLVM17 that need to be taken account of:

* ELF targets no longer do self-registration.
* The profile version has been raised to 9.

In addition we want the sample to visualize the profile with the run command so that we can more easily be alerted to changes in the profile format.

The largest change is the removal of self-registration. When this was in operation each object file has a .init_array entry with a pointer to a registration function that would be called prior to main. This registration function would register the names and counters of all the functions, allowing the program to calculate at run-time the extent of the counter and names section. Without self-registration we need to use
linker defined __start_ and __stop_ section symbols to find the extent.

The profile version increasing to 9 has added a bitmap section that we need to account for. This section should be empty unless MC/DC coverage is used.

The majority of the source code has been adapted from files in llvm-project/compiler-rt/lib/profile and
llvm-project/compiler-rt/include/profile

To replicate the behaviour of the previous example, we make a C __attribute__((constructor)) function that runs before main to register an atexit handler that writes out the profile after exiting main.


voltur01 (approved):
Thank you! Looks good to me and works like a charm!

PR #428, opened 2024-04-23
Shorten archive folder name for BMT main

vrukesh:


voltur01 (approved):
LGTM, just wonder if LLVM_TOOLCHAIN_PACKAGE_NAME is ever used, if we need this extra logic "just in case"?

PR #441, opened 2024-05-10
Add a way to set size vs performance optimization for library variants

voltur01:
We cannot add both code size and performance library sub-variants for every variant yet, because there is no way to select between them in the multilib file, so some defaults suggested for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -594,6 +594,7 @@ function(
     variant
     target_triple
     flags
+    build_type
     test_executor_params
     default_boot_flash_addr
     default_boot_flash_size
@@ -615,6 +616,12 @@ function(
         set(enable_long_double_test true)
     endif()
 
+    if(build_type MATCHES "release")
+        # Set additional performance options
+        set(newlib_nano_malloc "false")
+    else()
+        set(newlib_nano_malloc "true")
+    endif()
     ExternalProject_Add(
         picolibc_${variant}
         SOURCE_DIR ${picolibc_SOURCE_DIR}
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -615,6 +616,12 @@ function(
         set(enable_long_double_test true)
     endif()
 
+    if(build_type MATCHES "release")
+        # Set additional performance options
+        set(newlib_nano_malloc "false")
+    else()
```
smithp35, line 622: IIUC buildtype is a meson option https://mesonbuild.com/Builtin-options.html which can be one of plain, debug, debugoptimized, release, minsize, custom.

Would it be best to check for minsize to set newlib_nano_malloc to true, and everything else is the larger version.

CMakeLists.txt
```diff
@@ -1113,6 +1124,7 @@ function(add_library_variant target_arch)
         SUFFIX
         COMPILE_FLAGS
         MULTILIB_FLAGS
+        BUILD_TYPE
```
smithp35, line 1127: Would it be worth using `PICOLIBC_BUILD_TYPE` to make it more explicit that this is passing through --buildtype to picolibc?

smithp35 (commented):
I think it would be worth updating the README.md https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/blob/main/README.md to state that we build M profile libraries with size optimization and R and A with performance.

Possibly worth updating the source instructions at:
https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/blob/main/docs/building-from-source.md to indicate what a user would need to do to alter the pre

smithp35 (approved):
LGTM thanks for the updates

PR #442, opened 2024-05-10
Add check-all-embedded-toolchain target to group all checks

voltur01:


CMakeLists.txt
```diff
@@ -1897,3 +1897,14 @@ if(LLVM_TOOLCHAIN_CROSS_BUILD_MINGW)
...
 endif()
+
+add_custom_target(check-all-embedded-toolchain)
+
+add_dependencies(
+    check-all-embedded-toolchain
+    check-all
+    check-llvm-toolchain
+    check-package-llvm-toolchain
+    check-cxxabi
```
smithp35, line 1908: Is `check-cxx` missing? I can see `check-cxxabi` and `check-unwind` 

I would have intuitively expected `check-all` to include all the LLVM components including runtimes but I guess this isn't the case.

smithp35 (commented):
Is it worth updating the documentation with the new larger scope tests?

https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/blob/main/docs/building-from-source.md#testing-the-toolchain

smithp35 (approved):
Changes LGTM. Please can the title/description be updated prior to merging to match the implementation.

PR #451, opened 2024-07-18
Added Omax optimisation cfg file

nasherm:
Introduced a configuration file which enables several optimisation flags.

### DIFF
--- a/Omax.cfg
+++ b/Omax.cfg
@@ -1,0 +1,10 @@
+-O3 -ffast-math \
+-mllvm -inline-threshold=500 \
+-mllvm -unroll-threshold=450 \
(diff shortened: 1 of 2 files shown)


Omax.cfg
```diff
@@ -0,0 +1,10 @@
+-Ofast \
```
stuij, line 1: as `-Ofast` has now been depreciated and this now will emit a warning (https://github.com/llvm/llvm-project/pull/98736/files), I think it would make sense to change this to `-O3 -ffast-math`.

nasherm: Done!

stuij (approved):
Thanks! LGTM.

PR #452, opened 2024-07-22
Add experimental ability to build LLVM libc

statham-arm:
LLVM libc is now complete enough that you can compile it in "full library" mode for AArch32 bare metal and run a hello-world program (and surely many other things too, but that's all I've tried so far). However, setting it up is not trivial, and takes effort. To make it easier for users to evaluate it in the context of this toolchain, this patch series adds LLVM libc support to the main cmake file, plus a small collection of porting and startup functions.


voltur01 (approved):
Thank you for progressing this! The change looks good to me and works as well.

Two things that we should address for the overlay packages as a follow up:
* Licenses - newlib package seems to have copies of its licenses in addition to the main project THIRD-PARTY-LICENSES.txt, while llvmlibc one does not have the copy of https://github.com/llvm/llvm-project/blob/main/libc/LICENSE.TXT
* We need to find a way to capture the upstream version, the full package provides VERSION.txt

PR #455, opened 2024-07-23
When building LLVM libc, include its license file.

statham-arm:
This copies `libc/LICENSE.TXT` from the llvm-project repository into the `third-party-licenses` directory, renamed to `LIBC-LICENSE.TXT` so you can tell it apart from all the other subdirectory licenses. (We were already doing similar renames for libcxx, libunwind, etc.)

When you build the toolchain in llvm-libc-only mode, that file shows up along with all the others at the top level of `third-party-licenses`. If you build an overlay package, it ends up in the `third-party-licenses/llvmlibc` subdirectory.


voltur01 (approved):
Thank you for addressing this promptly!

PR #456, opened 2024-07-25
Separate LTO options from Omax.cfg

nasherm:
The motivation for this change is that users
may wish to disable LTO when using the the Omax.cfg
file. Removing the LTO flags to a seperate file allows
the user to enable LTO related flags seperately.


stuij (commented):
LGTM

stuij:
Oh right, this repo doesn't do auto-squashing. Next time we should make sure we don't have spurious commits in the branch before merging.

PR #458, opened 2024-08-01
Remove patch since llvm sort.pass.cpp is already marked as a long test

vrukesh:
Patch https://github.com/llvm/llvm-project/pull/100720 makes this patch redundant: 0004-libc-tests-with-picolibc-mark-sort-test-as-long-one.patch
So, removed it and renamed other patches


ostannard (approved):
LGTM

PR #461, opened 2024-08-08
Update README to reference Omax.cfg and OmaxLTO.cfg

nasherm:
Updating the README.md file to detail use case of Omax.cfg and OmaxLTO.cfg.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,29 @@ $ clang \
 The FPU selection can be skipped, but it is not recommended to as the defaults
 are different to GCC ones.
 
+
+The builds of the toolchain come packaged with two config files, Omax.cfg and OmaxLTO.cfg.
+When used, these config files enable several build optimisation flags to achieve highest performance on typical embedded benchmarks. OmaxLTO.cfg enables link-time optimisation (LTO) specific flags.
+These configs can be optionally passed using the `--config` flag. For example:
+
+```
+$ clang \
+example.c \
+...
+--config=Omax.cfg \
+--config=OmaxLTO.cfg \
+-o example
+```
+
+Users should be warned that Omax.cfg enables `-ffast-math` which breaks IEEE compliance and
+enables maths optimisations which can affect code correctness.  LTOs are
+kept seperately in OmaxLTO.cfg as users may not want LTOs due to potential increase in link time
+and/or increased memory usage during linking. The flags used in these configs are used internally by
+us at Arm, information on what they do can be found at
+[llvm.org](https://clang.llvm.org/docs/CommandGuide/clang.html).
+Users are also encouraged to create their own configs and tune their own flag parameters.
+
+
 Binary releases of the LLVM Embedded Toolchain for Arm are based on release
 branches of the upstream LLVM Project, thus can safely be used with all tools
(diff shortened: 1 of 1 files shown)


stuij (commented):
I feel we should warn the reader that the optimisations include `-ffast-math`, which breaks IEEE compliance, amongst others.

README.md
```diff
@@ -140,6 +140,19 @@ $ clang \
 -o example example.c
 ```
 
+The builds of the toolchain come packaged with two config files, Omax.cfg and OmaxLTO.cfg.
+When used, these config files enable several build optimisation flags. OmaxLTO.cfg enables LTO
```
voltur01, line 144: Would be nice to explain:
...optimization flags to achieve highest performance on typical embedded benchmarks.

voltur01, line 144: Any recommendation when to use them (LTO and no LTO) together or separately?

voltur01, line 144: Would be also nice to mention that the flags used are internal ones, any advice where to look to explanation of what these options are doing?

README.md
```diff
@@ -140,6 +140,19 @@ $ clang \
...
+
+```
+$ clang \
+example.c \
+...
+--config=Omax.cfg \
+--config=OmaxLTO.cfg \
+-o example
+```
+
```
voltur01, line 155: Could we also add a hint like it is possible to use these config files as an inspiration to create config file for your project by picking options from there and tuning the parameters as fits the project?

voltur01 (commented):
Thank you for adding the details! Please see some ideas how to expand it a bit.

README.md
nasherm: Done!

nasherm: I've updated and provided a link to clang documentation in my newest patch.

nasherm: I've made reference to how LTO can potentially increase link time and memory usage during linking

README.md
nasherm: Done!

nasherm:
> I feel we should warn the reader that the optimisations include `-ffast-math`, which breaks IEEE compliance, amongst othe

@stuij Provided a brief explanation of the risk of `-ffast-math` in my newest patch

README.md
```diff
@@ -143,6 +143,28 @@ $ clang \
...
+Users should be warned that Omax.cfg enables `-ffast-math` which breaks IEEE compliance and
+enables maths optimisations which can affect code correctness.  LTOs are
+kept separately in OmaxLTO.cfg as users may not want LTOs due to potential increase in link time
+and/or increased memory usage during linking. The flags used in these configs are used internally by
+us at Arm, information on what they do can be found at
```
voltur01, line 164: Thank you for the update! Sorry, I was not clear: I meant that these are clang internal options (-mllvm ones) and I believe they are not documented - best I was able to find for some of them was comments in the source code. Maybe this is the way to put it: config files use llvm internal undocumented options, please see source code of corresponding optimization passes in LLVM project to find out more.

nasherm: Ah, I see. Updated

voltur01 (approved):
Looks good to me, thank you for the quick turnaround!

PR #462, opened 2024-08-13
Support building LLVM libc on AArch64 (plus minor fixes)

statham-arm:
The first two commits in this PR arrange to build the experimental LLVM libc on AArch64 as well as AArch32 (fixing a build failure and then enabling the build).

The other two are minor cleanups, developed together but not strongly related, but too small to bother making a separate PR for.


ostannard (approved):
LGTM

PR #464, opened 2024-08-20
Add missing armv7 library variants.

simpal01:
This patch adds the following library variants.

                1. armv7r hard_vfpv3xd
                2. armv7a soft_vfpv3_d16.

Also disable the math_errhandling tests to the set of disabled
tests for the armv7r_hard_vfpv3xd library variant.These tests were
already disabled for the previously existing variants, due to a known
issue in compiler-rt where the floating point exceptions are not set
correctly for computations on types implemented in software.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1605,7 +1605,7 @@ add_library_variants_for_cpu(
 )
 add_library_variants_for_cpu(
     armv7a
-    SUFFIX softfp_vfpv3_d16
+    SUFFIX soft_vfpv3_d16
     COMPILE_FLAGS "-mfloat-abi=softfp -march=armv7a -mfpu=vfpv3-d16"
     MULTILIB_FLAGS "--target=armv7-unknown-none-eabi -mfpu=vfpv3-d16"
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -1636,6 +1653,23 @@ add_library_variants_for_cpu(
     RAM_SIZE 0x1000000
     STACK_SIZE 4K
 )
+add_library_variants_for_cpu(
+    armv7r
+    SUFFIX soft_vfpv3_d16
+    COMPILE_FLAGS "-mfloat-abi=soft -march=armv7r -mfpu=vfpv3-d16"
```
ostannard, line 1659: These options contradict one another - `-mfloat-abi=soft` disables the FPU, but then `-mfpu=vfpv3-d16` enables it. Should this be `-mfloat-abi=softfp` to match the other library variant, or something else?

simpal01: Done

simpal01:
> This should also have tests in `test/multilib/armv7a.test` to make sure that they get selected correctly.

Added the tests.

ostannard (approved):
LGTM

PR #468, opened 2024-08-22
Make aarch64-a variant distinct from aarch64-r

eleanor-arm:
Currently there's only one aarch64-a variant. We would like to add
aarch64-r variants in the future and need a way of distinguishing
between the different profiles.


ostannard (approved):
LGTM

Issue #470, opened 2024-08-25
Relpath broken in custom build guide


itxs:
![image](https://github.com/user-attachments/assets/a40bb05c-1883-4e17-a82f-9098a1efe4b2)

Paths with "$PWD"/../xxx should be actually just "$PWD"/xxx

stuij:
fixed by https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/pull/472

PR #471, opened 2024-08-25
Fixed command line example

itxs:


stuij:
The relative path issue was fixed by https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/pull/472.

Sorry I just saw an email of the issue you raised (https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/issues/470), and had a local fix that I forgot to upstream, so I figured I'd make a quick patch. I didn't notice you made a patch as well.

docs/building-from-source.md
```diff
@@ -61,11 +63,13 @@ If you prefer you can check out and patch the repos manually and use those.
 If you check out repos manually then it is your responsibility to ensure that the correct revisions are checked out - see `versions.json` to identify these.
 
 ```
+export CC=clang
+export CXX=clang++
 mkdir repos
-git -C repos clone https://github.com/llvm/llvm-project.git
```
stuij, line 65: For me, the point of specifying an alternate place for the repos is because I want to develop on the code base. And so I want to have the git history.

Otherwise you can just not specify it, and it will be downloaded for you without the history by CMake.

itxs: Readme file is called building-from-source, and it will be used more often only for build, rather than for developing. One who wants to develop is able to perform the same command manually, but without "--depth=1". After all, the development is much rarer than building

docs/building-from-source.md
```diff
@@ -61,11 +63,13 @@ If you prefer you can check out and patch the repos manually and use those.
...
-git -C repos/llvm-project am -k "$PWD"/../patches/llvm-project/*.patch
```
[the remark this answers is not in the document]
itxs: Done

docs/building-from-source.md
itxs: Ok, I removed the "--depth=1"

PR #472, opened 2024-08-28
Fix path in build from source recipe

stuij:
In the `Building` section of building-from-source.md, twice we are trying to apply a patch with the following command:

    git -C repos/<repo> am -k "$PWD"/../patches/<repo>/*.patch

This command is executed from the root of the repo. The original thought for this must have been that because of the `-C`, our path at time of execution is `repos/<repo>`, and thus we need to move a level up from our current directory. However `$PWD` will be replaced by the shell before the command is executed. So we need to remove the spurious `../`.


ostannard (approved):
LGTM

PR #473, opened 2024-08-30
Add missing set of disabled tests for PACBTI-M variants

pratlucas:
This adds the `fenv` and `math_errhandling` tests to the set of disabled
tests for the PACBTI-M library variants, introduced by #467.

These tests were already disabled for the previously existing variants,
due to a known issue in compiler-rt where the floating point exceptions
are not set correctly for computations on types implemented in software.


ostannard (approved):
LGTM

PR #479, opened 2024-09-02
Add multilib mappings for thumb triples

ostannard:
The instruction set (ARM/Thumb) is included in the triple, but doesn't matter for library selection because the ISA can be switched on any function call. This adds match blocks to multilib.yaml to treat thumb triples as if they were arm triples for library selection.

This will also be needed for AArch32 v8-A and v8-R, but we don't have any libraries for them yet.


voltur01 (approved):
LGTM

PR #480, opened 2024-09-03
Update a downstream patch for altered context (NFC)

statham-arm:
This patch failed to apply for me in a manual `git am`, because the upstream addition of `ADDITIONAL_COMPILE_FLAGS` was just close enough to our change to confuse git. Regenerated that hunk. Now it still makes the same change, just without complaining.


smithp35 (approved):
LGTM.

PR #482, opened 2024-09-03
Merge v7-M and v7E-M library variants

ostannard:
The only differece between v7-M and v7E-M is the DSP extension, which is not useful for the C/C++ libraries, so we can simplify things by just having v7-M builds. This fills in a gap of (v7-M, fpv4-sp-d16, hard-float), which we did not previously have a valid library for.


statham-arm (approved):
There's one instruction in v7E-M which _is_ potentially useful to non-DSP-focused runtime libraries, and that's UMAAL, the form of 32×32→64 bit multiply+add that takes two 32-bit addends instead of one 64-bit one. That's much better than UMLAL for multiprecision arithmetic, and can speed up double-precision FP multiply as well (if you're doing that without an FPU).

But I agree that at the moment it doesn't demand a whole extra set of libraries (especially since as far as I know compiler-rt doesn't provide an optimized DP multiply that uses UMAAL anyway).

PR #483, opened 2024-09-03
Statically link libc++abi into libc++

ostannard:
This will prevent C++ links from breaking when
https://github.com/llvm/llvm-project/pull/101259 lands, as that removes -lc++abi and -lunwind from the linker command-line.


smithp35 (approved):
LGTM thanks for making the change.

PR #487, opened 2024-09-05
Fixing the multilib tests introduced by #464

simpal01:
Use a different check prefix for the newly added tests.


ostannard (approved):
LGTM

PR #490, opened 2024-09-05
Reorder the add_library_variant calls for the armv7r_hard_vfpv3_d16 to win

simpal01:
The order in multilib.yaml matches the order of the libraries in CMakeLists.txt.

At the moment both the below commandline options selecting the same library variant.
```
$ clang -print-multi-directory --target=armv7r-none-eabihf -mfpu=vfpv3-d16
arm-none-eabi/armv7r_hard_vfpv3xd_exn_rtti
$ clang -print-multi-directory --target=armv7r-none-eabihf -mfpu=vfpv3xd
arm-none-eabi/armv7r_hard_vfpv3xd_exn_rtti
``` 
We need to reorder the the add_library_variant calls so that armv7r_hard_vfpv3_d16 gets selected appropriately.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1639,9 +1639,9 @@ add_library_variants_for_cpu(
 )
 add_library_variants_for_cpu(
     armv7r
-    SUFFIX hard_vfpv3_d16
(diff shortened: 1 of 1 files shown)


ostannard (approved):
LGTM

PR #491, opened 2024-09-06
Add missing armv7 library variants

simpal01:
This patch adds the following library variant.

     1. armv7r soft_vfpv3_d16.

Adding tests in test/multilib to make sure that all the newly added library variants  get selected correctly..


ostannard (approved):
LGTM

PR #493, opened 2024-09-09
Add missing armv7 library variants. (#464)

simpal01:
This patch adds the following library variants.

            1. armv7r hard_vfpv3xd
            2. armv7a soft_vfpv3_d16.

Also disable the math_errhandling tests to the set of disabled tests for the armv7r_hard_vfpv3xd library variant.These tests were already disabled for the previously existing variants, due to a known issue in compiler-rt where the floating point exceptions are not set correctly for computations on types implemented in software.

Adding tests in test/multilib to make sure that all the newly added library variants  get selected correctly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1639,9 +1639,9 @@ add_library_variants_for_cpu(
 )
 add_library_variants_for_cpu(
     armv7r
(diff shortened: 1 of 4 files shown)


dcandler (approved):
LGTM

PR #495, opened 2024-09-10
Add LoopFullUnroll pass to the LTO pipeline.

VladiKrapp-Arm:
This addition improves performance in some scenarios.


ostannard:
This looks like something which should be committed to LLVM directly, we'd rather avoid carrying patches of our own if possible.

PR #496, opened 2024-09-10
Various fixes for the LLVM 19 branch

dcandler:
There were a number of changes to the toolchain that happened after the LLVM 19 branch that can be reverted.
Likewise, the LLVM 19 branch is lacking certain commits which need to be cherry picked.
There is also a minor issue with the LLVM libc namespace on release branches, which requires a workaround.


ostannard (approved):
LGTM

PR #497, opened 2024-09-12
Create a custom config for LLVM libc

dcandler:
Since LLVM libc lacks an AArch64 config, LIBC_CONFIG_PATH is set to the arm subdirectory to force it to use the existing AArch32 config. However, this has the side effect of never loading the generic baremetal config.json file, which is instead in the parent directory.

In order to combine the entrypoints in the arm config with the values set in the generic baremetal config, this patch copies both to the same location to be used in the build.


pratlucas (approved):
LGTM.

PR #498, opened 2024-09-12
Enable PACBTI in picolibc's startup code

vhscampos:
Follow up to [#467](https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/pull/467).

This patch makes picolibc's startup code enable PAC and BTI in the relevant library variants.

picolibc's startup code enables PAC and BTI if and only if __ARM_FEATURE_PAUTH and __ARM_FEATURE_BTI are defined respectively. These macros are defined if `+pacbti` is passed in the `-march` command-line option: it is not enough to use `-mbranch-protection` in the build of the library variant.


pratlucas (approved):
LGTM. Thanks for catching this!

PR #500, opened 2024-09-12
Diagnose an unsupported multilib setup: MVE soft-float.

statham-arm:
We don't build any MVE-capable libraries with the soft-float ABI. We do have Armv7-M soft-float libraries, but you can't fall back to using those, because MVE requires the FPSCR to be set up specially at startup time, and our v7-M soft-float libraries don't do that.

So, rather than select one of those libraries and silently generate wrong MVE code, we should force an error report instead. This commit arranges that, by using the new FatalError flag recently added to the multilib.yaml mechanism.

We already had a commented-out test for this case, so I've uncommented it and made it expect a compile failure with the custom error message.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1441,6 +1441,26 @@ function(add_library_variant target_arch)
     set(multilib_yaml_content "${multilib_yaml_content}" PARENT_SCOPE)
 endfunction()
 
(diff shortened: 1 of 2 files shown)


voltur01 (approved):
Makes sense to me, thank you!

PR #504, opened 2024-09-18
Adjust clang's multilib flags test to cope with missing variant

pratlucas:
The clang driver's `print-multi-selection-flags.c` test uses a
combination of command line options in one of its RUN lines for which we
don't have a valid library variant available.
Since the merging of PR #500, this combination of options now trigger an
error to inform the user of an unsupported setup, but this error also
causes the test to fail.
This adds a new patch file for llvm-project to adjust the test
accordingly.


ostannard:
Would it instead be possible to fix the test upstream to avoid using our multilib.yaml at all? There are other tests, e.g. `baremetal-multilib.yaml`, which do that by building a directory tree containing symlinks to the compiler and multilib.yaml under test, and some dummy library files. It might be cleaner for this test to add a driver option to use a different multilib.yaml, but I suppose the other tests need that whole tree including the libraries, so that might not be worth it.

pratlucas:
The change to upstream is under review, but it's taking a while to progress. Should we go ahead with this change for now and revert it once the upstream PR gets merged?

ostannard (approved):
LGTM as a temporary workaround.

PR #506, opened 2024-09-18
Use FVP models to test v8.1-M libraries with PACBTI

ostannard:
QEMU doesn't currently support the M-profile PACBTI extension, so we need to use FVP models to test those libraries. There are also other targets not supported by QEMU, such as v8-R, which we will need these for.

These models are available free of cost (though not open source) from Arm, and the user must agree to a EULA before using them. I've added a script `fvp/get_fvps.sh' to download and extract them into the `fvp' directory.

These models are currently only used for the v8.1-M libraries with PACBTI, but the script also downloads the AEMv8A and AEMv8R models for future use. If the FVPs aren't installed, then the CMakeLists.txt will skip any tests which need them. I don't think it's worth preserving the ability to run these tests with QEMU, because we still have the non-PACBTI library variants which are run with QEMU. The two models also have different memory maps and require different scatter files, so if we wanted the ability to run tests for one library on both then we'd need to add a way to build multiple linker scripts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,14 @@ __pycache__
 # IDEs
 .vscode
 compile_commands.json
+
+# Downloaded FVPs
+/fvp/AEMv8R_base_pkg/
+/fvp/Base_RevC_AEMvA_pkg/
+/fvp/Corstone-310/
+/fvp/FVP_Corstone_SSE-310.sh
+/fvp/FastModelsPortfolio_11.26/
+/fvp/FastModels_crypto_11.26.011_Linux64/
+/fvp/complete
+/fvp/download/
+/fvp/license_terms/
--- a/test-support/lit-exec-fvp.py
+++ b/test-support/lit-exec-fvp.py
@@ -1,0 +1,90 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2023-2024, Arm Limited and affiliates.
+
+# This script is a bridge between lit-based tests of LLVM C++ runtime libraries
+# (libc++abi, libunwind, libc++) and FVP models. It must handle the same
+# command-line arguments as llvm-project/libcxx/utils/run.py.
+
+from run_fvp import run_fvp
+import argparse
+import pathlib
+import sys
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Run a single test using qemu"
+    )
+    parser.add_argument(
+        "--fvp-root-dir",
+        help="Directory in which FVP models are installed",
+        required=True,
+    )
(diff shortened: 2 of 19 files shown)


.gitignore
```diff
@@ -11,3 +11,14 @@ __pycache__
 # IDEs
 .vscode
 compile_commands.json
+
+# Downloaded FVPs
+/fvp/AEMv8R_base_pkg/
```
pratlucas, line 16: To make this easier to maintain in the long run, would it better to move the FVP's install directory to live inside the build directory instead?
That way we decouple it from the source tree, and make it simpler for developers to use their own out-of-tree FVP installations if they wish to.

test-support/lit-exec-fvp.py
```diff
@@ -0,0 +1,90 @@
...
+
+from run_fvp import run_fvp
+import argparse
+import pathlib
+import sys
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Run a single test using qemu"
```
voltur01, line 17:
```suggestion
        description="Run a single test using Arm FVP"
```

voltur01 (commented):
Thanks, looks good to me, especially the docs part, however I think I spotted a place where qemu needs to be updated to FVP.

.gitignore
ostannard: There is already an `FVP_INSTALL_DIR` cmake option to point it at an existing copy of the FVPs, but I've just realised that I'm also accessing the config files using that, so I'll re-arrange things to have two directories. However, I think leaving it in the source tree by default makes it simpler to get running, because `get_fvps.sh` knows where the source directory is, but not the install directory.

pratlucas: Thanks! In that case, could we have a new directory under `fvp` to hold all of the downloaded FVPs (e.g. `fvp/install/`)? That way we can gitignore this entire directory and don't have to maintain this list of individual FVPs going forwards.

ostannard: That's exactly what the latest patch does:
* `fvp/download/` holds the downloaded files, is only used by `get_fvps.sh`, and is in .gitignore
* `fvp/install/` holds the installed FVPs, this can be moved using the `FVP_INSTALL_DIR` cmake option, and is in .gitignore
* `fvp/config/` holds the config files, and is normal source code, not in .gitignore, and accessed relative to `CMAKE_CURRENT_SOURCE_DIR` so it doesn't need to be copied to an external FVP installation

pratlucas: Oh, brilliant! Thanks! Sorry I missed the patch updates in the PR.

PR #508, opened 2024-09-18
Reference A-profile correctly [NFC].

smithp35:
Add missing dash between A and profile to make A-profile.


voltur01 (approved):
Thanks for fixing this!

PR #509, opened 2024-09-20
Build libraries even when FVPs not installed

ostannard:
This was the intention of the previous patch, but I put the check in the wrong place, so the actual builds of the library variants which need FVPs were disabled, not just the tests.

This doesn't need passing through to the add_newlib or add_llvmlibc functions because we don't run the tests for them yet.


dcandler (approved):
LGTM, thanks.

PR #511, opened 2024-09-30
Add llvm-strings to the distribution

ostannard:

PR #513, opened 2024-09-30
Add llvm-strings and clang-cpp to the distribution

ostannard:
This is a cherry-pick of #511 and #512 to the llvm-19 branch.


voltur01 (approved):
Thank you!

PR #518, opened 2024-10-03
Add additional patch folder for performance patches

dcandler:
This adds a second llvm-project patch directory, intended for performance enhancement patches which can be kept optional to avoid blocking builds if they become out of date with the main branch. A new CMake option controls whether or not to apply the patches, and a new script is included to manage the patching process to make it easier to identify when there is a problem with a patch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -134,6 +134,11 @@ option(
     the tools every time you update llvm-project."
     ON
 )
+option(
+    APPLY_LLVM_PERFORMANCE_PATCHES
+    "During checkout, apply optional downstream patches to
+    llvm-project to improve performance."
+)
 set(
     FVP_INSTALL_DIR
     "${CMAKE_CURRENT_SOURCE_DIR}/fvp/install" CACHE STRING
@@ -280,24 +285,17 @@ if(NOT (LLVM_TOOLCHAIN_C_LIBRARY STREQUAL llvmlibc)) # libc in a separate repo?
   read_repo_version(${LLVM_TOOLCHAIN_C_LIBRARY} ${LLVM_TOOLCHAIN_C_LIBRARY})
 endif()
 
-# The patches are generated from custom branch, with followin command:
-#   git format-patch -k origin/main
-# The patches apply cleanly against llvm-project commit
-# 327124ece7d59de56ca0f9faa2cd82af68c011b9.
-set(
-    llvm_project_patches
-    ${CMAKE_CURRENT_SOURCE_DIR}/patches/llvm-project/0001-libc-tests-with-picolibc-xfail-one-remaining-test.patch
-    ${CMAKE_CURRENT_SOURCE_DIR}/patches/llvm-project/0002-libc-tests-with-picolibc-disable-large-tests.patch
-    ${CMAKE_CURRENT_SOURCE_DIR}/patches/llvm-project/0003-Disable-failing-compiler-rt-test.patch
-    ${CMAKE_CURRENT_SOURCE_DIR}/patches/llvm-project/0004-libc-tests-with-picolibc-XFAIL-uses-of-atomics.patch
-    ${CMAKE_CURRENT_SOURCE_DIR}/patches/llvm-project/0005-libc-tests-with-picolibc-mark-two-more-large-tests.patch
-)
+set(LLVM_PATCH_COMMAND ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/cmake/patch_llvm.py ${CMAKE_CURRENT_SOURCE_DIR}/patches/llvm-project)
+if(APPLY_LLVM_PERFORMANCE_PATCHES)
+    set(LLVM_PATCH_COMMAND ${LLVM_PATCH_COMMAND} && ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/cmake/patch_llvm.py ${CMAKE_CURRENT_SOURCE_DIR}/patches/llvm-project-perf)
+endif()
+
 FetchContent_Declare(llvmproject
     GIT_REPOSITORY https://github.com/llvm/llvm-project.git
     GIT_TAG "${llvmproject_TAG}"
     GIT_SHALLOW "${llvmproject_SHALLOW}"
     GIT_PROGRESS TRUE
-    PATCH_COMMAND git reset --quiet --hard && git clean --quiet --force -dx && git am -k --ignore-whitespace --3way ${llvm_project_patches}
+    PATCH_COMMAND ${LLVM_PATCH_COMMAND}
     # Add the llvm subdirectory later to ensure that
     # LLVMEmbeddedToolchainForArm is the first project declared.
     # Otherwise CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT
--- a/cmake/patch_llvm.py
+++ b/cmake/patch_llvm.py
@@ -69,7 +69,7 @@ def main():
             sys.exit(0)
         if args.restore_on_fail:
             # Check that the operation can be aborted.
-            # git am does give any specific return codes,
+            # git am doesn't give any specific return codes,
             # so check for unresolved working files.
             rebase_apply_path = os.path.join(".git", "rebase-apply")
             if args.llvm_dir:
--- a/cmake/patch_llvm.py
+++ b/cmake/patch_llvm.py
@@ -28,7 +28,7 @@ def main():
     )
     parser.add_argument(
         "--reset",
-        help="Clean and reset the repo to a specified commit before patching.",
+        help="Clean and hard reset the repo to a specified commit before patching.",
     )
     parser.add_argument(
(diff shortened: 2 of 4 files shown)


cmake/patch_llvm.py
```diff
@@ -0,0 +1,132 @@
...
+    if args.reset:
+        reset_args = ["git", "reset", "--quiet", "--hard"]
+        subprocess.check_output(reset_args, cwd=args.llvm_dir)
+        clean_args = ["git", "clean", "--quiet", "--force", "-dx"]
+        subprocess.check_output(clean_args, cwd=args.llvm_dir)
+
+    patch_names = []
+    for patch_name in os.listdir(args.patchdir):
+        if patch_name.endswith(".patch"):
+            patch_names.append(patch_name)
```
pratlucas, line 49: This can be simplified using list comprehension:
```
patch_names = [patch for patch in os.listdir(args.patchdir) if patch.ends_with(".patch")]
```

cmake/patch_llvm.py
```diff
@@ -0,0 +1,132 @@
...
+    )
+    parser.add_argument(
+        "--method",
+        choices=["am", "apply"],
+        default="apply",
+        help="Git command to use. git am will add each patch as a commit, whereas git apply will leave patched changes staged.",
+    )
+    parser.add_argument(
+        "--reset",
+        help="Clean and reset the repo to a specified commit before patching.",
```
pratlucas, line 31: The help message says the option should reset the repo to a "specified commit", but the logic for it resets the repo to `HEAD`.

cmake/patch_llvm.py
```diff
@@ -0,0 +1,132 @@
...
+
+    patch_names = []
+    for patch_name in os.listdir(args.patchdir):
+        if patch_name.endswith(".patch"):
+            patch_names.append(patch_name)
+    patch_names.sort()
+
+    print(f"Found {len(patch_names)} patches to apply:")
+    for patch_name in patch_names:
+        print(patch_name)
```
pratlucas, line 54: Nit: `print(patch_names.join("\n"))`

cmake/patch_llvm.py
```diff
@@ -0,0 +1,132 @@
...
+
+    if args.method == "am":
+        merge_args = ["git", "am", "-k", "--ignore-whitespace", "--3way"]
+        for patch_name in patch_names:
+            merge_args.append(os.path.join(args.patchdir, patch_name))
```
pratlucas, line 59: This can also make use of list comprehension:
```
merge_args += [os.path.join(args.patchdir, patch) for patch in patch_names]
```

cmake/patch_llvm.py
```diff
@@ -0,0 +1,132 @@
...
+        print(p.stderr)
+
+        if p.returncode == 0:
+            print(f"All patches applied.")
+            sys.exit(0)
+        if args.restore_on_fail:
+            # Check that the operation can be aborted.
+            if (
+                'To restore the original branch and stop patching, run "git am --abort".'
+                in p.stdout
```
pratlucas, line 73: This string matching might be an unstable solution, as there's no guarantee this message wont change between git versions.
In which scenarios running `git am --abort` wouldn't be acceptable after a an error code? Can they be detected in a different way?

pratlucas (commented):
Small suggestion: should this repo patching infrastructure deal with the possibility of local changes being present in the checkouts, perhaps by giving up and failing if that's the case?
The repositories can end up quite messy if local changes exist, especially when applying the patches with `git apply`.

CMakeLists.txt
```diff
@@ -134,6 +134,11 @@ option(
     the tools every time you update llvm-project."
     ON
 )
+option(
+    APPLY_LLVM_PERFORMANCE_PATCHES
+    "During checkout, apply optional downstream patches to
+    llvm-project to improve performance."
+)
```
VladiKrapp-Arm, line 141: This option should probably be described in the comment block on top of the file. 

Should this really be off by default? Users would probably want to get perf improvements by default.

cmake/patch_llvm.py
VladiKrapp-Arm, line 49: Here, and onwards: 
The code currently extracts the file names only to later append them to the folder name and use that, multiple times. 
What we actually want is to work with file paths: 
Use glob to get all the relevant files, as we'd simply do in the shell : *.patch. 
Then we use file paths everywhere, and in the single place we just want the names, we simply do path.name().stem(). 

If I understand correctly, using pathlib 
pathlib.Path(args.patchdir).glob("*.patch")
would return an iterable with the relevant paths.

cmake/patch_llvm.py
```diff
@@ -0,0 +1,132 @@
...
+                            "git",
+                            "apply",
+                            "--ignore-whitespace",
+                            "--3way",
+                            "--reverse",
+                            r_patch,
+                        ]
+                        p_check = subprocess.run(
+                            reverse_args, cwd=args.llvm_dir, check=True
+                        )
```
VladiKrapp-Arm, line 125: Would it not be simpler to just remember the commit before we started applying patches, and do a git reset to that if a patch fails? One liner, and no need to keep track of applied patches.

cmake/patch_llvm.py
```diff
@@ -0,0 +1,132 @@
...
+import os
+import subprocess
+import sys
+
+
+def main():
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "patchdir",
+        help="Set of patches to apply. This should be a directory containing one or more ordered *.patch files.",
```
VladiKrapp-Arm, line 17: The code as currently written requires this to be provided as an absolute path. We should either ease this requirement, or document it here. 

I'm considering the scenario where the user wants to run this script to apply patches manually. The scenario as described in CMakeLists.txt has them in the root of LLVM-embedded-toolchain-for-Arm.

CMakeLists.txt
dcandler: The aim is that the default configuration should work out of the box. Downstream changes are additional points of failure (we would like to remove patches we already have), so I would prefer to prioritize the toolchain building first.

Release branches might be a different matter, since they use a defined and therefore stable version of LLVM.

cmake/patch_llvm.py
dcandler: I can see how the paths can get muddled since the git commands need to run in the llvm-project location, whereas the patch paths will be specified relative to the script. So I've adjusted the git arguments to use `-C` and convert the patch paths to absolute automatically.

cmake/patch_llvm.py
dcandler: Thanks, done.

cmake/patch_llvm.py
dcandler: Thanks, done.

cmake/patch_llvm.py
dcandler: A reset would be simple, but would also reset all changes, including any other patch sets that were previously applied or any local changes by the user. Rewinding is more involved, but gets you back to the same state as you started with.

cmake/patch_llvm.py
dcandler: Added the missing arg.

dcandler:
> Small suggestion: should this repo patching infrastructure deal with the possibility of local changes being present in the checkouts, perhaps by giving up and failing if that's the case?
> The repositories can end up quite messy if local changes exist, especially when applying the patches with `git apply`.

The automatic patching currently only occurs on a new checkout when the dependencies are initially fetched, so there shouldn't be any existing changes to conflict with at that time. When working with a pre-existing (and potentially modified) checkout, it's left to the user to handle the patching.

ostannard (commented):
Should we switch the `picolibc` and `newlib` patches to use this new script too, so that they work the same way? Currently it's slightly annoying that picolibc is applied as uncommitted changes, since I then have to commit them if I want to make any other changes to picolibc.

dcandler:
> Should we switch the `picolibc` and `newlib` patches to use this new script too, so that they work the same way? Currently it's slightly annoying that picolibc is applied as uncommitted changes, since I then have to commit them if I want to make any other changes to picolibc.

I already had in mind a followup patch which expands the use of the script, including adding a CMake option to select between `git am` and `git apply` (which is already an option on the patch script, but not selectable in the CMake project). I can incorporate the other repos at that point too.

cmake/patch_llvm.py
dcandler: Done, working with path objects does work out cleaner.

cmake/patch_llvm.py
```diff
@@ -0,0 +1,132 @@
...
+        p = subprocess.run(merge_args, capture_output=True, text=True)
+        print(p.stdout)
+        print(p.stderr)
+
+        if p.returncode == 0:
+            print(f"All patches applied.")
+            sys.exit(0)
+        if args.restore_on_fail:
+            # Check that the operation can be aborted.
+            # git am does give any specific return codes,
```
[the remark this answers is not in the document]
dcandler: Done

pratlucas (approved):
LGTM.

PR #521, opened 2024-10-04
Update llvmlibc.md documentation to include AArch64 support and heap.

smithp35:
Support for AArch64 has been available for AArch64 in llvm-libc for some time, but the docs still claim this isn't supported.

Also update the instructions for defining the bounds of the heap for when malloc is used.


voltur01 (approved):
Good catch, thank you!

PR #523, opened 2024-10-05
Add aarch64a_soft_nofp variant

keith-packard:
Build a soft float multilib variant for aarch64 supporting targets without an FPU.

This contains a couple of minor fixes for llvm to enable this; I'll be submitting those upstream.


patches/llvm-project/0007-libunwind-Support-aarch64-without-FPU.patch
```diff
@@ -0,0 +1,190 @@
...
+ DEFINE_LIBUNWIND_FUNCTION(__libunwind_Registers_arm64_jumpto)
+   // skip restore of x0,x1 for now
+-  ldp    x2, x3,  [x0, #0x010]
+-  ldp    x4, x5,  [x0, #0x020]
+-  ldp    x6, x7,  [x0, #0x030]
+-  ldp    x8, x9,  [x0, #0x040]
+-  ldp    x10,x11, [x0, #0x050]
+-  ldp    x12,x13, [x0, #0x060]
+-  ldp    x14,x15, [x0, #0x070]
++  LDP(x2, x3, x0, #0x010, #0x018)
```
ostannard, line 43: The LDP instructions with X registers should be available without the FPU, is this part of the change needed?

ostannard (commented):
For the LLVM changes, we're trying to minimise the side of our downstream patch, so we'd prefer to wait until these have been committed to LLVM itself.

For the testing, have you checked if QEMU correctly models the undefined instruction exception if a floating-point instruction is executed in this library variant? I did some testing a while ago and couldn't get that to happen, which is why I'm trying to make FVPs usable for testing, since they can accurately model a lot of less-common architectures like this.

patches/llvm-project/0007-libunwind-Support-aarch64-without-FPU.patch
keith-packard: Nope. I'm not sure why I thought that was necessary. I'll be updating the upstream PR once I've finish testing.

keith-packard:
> For the LLVM changes, we're trying to minimise the side of our downstream patch, so we'd prefer to wait until these have been committed to LLVM itself.

Makes sense. I'm working on getting those upstream.

> For the testing, have you checked if QEMU correctly models the undefined instruction exception if a floating-point instruction is executed in this library variant? I did some testing a while ago and couldn't get that to happen, which is why I'm trying to make FVPs usable for testing, since they can accurately model a lot of less-common architectures like this.

I don't think QEMU has a model for aarch64 without an FPU, so all of my testing uses the full machine. However, my test harness leaves the FPU disabled when testing the nofp code. When it hits an instruction using the FPU, qemu traps.

keith-packard:
It would be good to have these changes tested on FVP; I haven't figured out how to do that yet.

keith-packard:
> Hi @keith-packard , Now that [llvm/llvm-project#111235](https://github.com/llvm/llvm-project/pull/111235) has been merged, I believe this change can go ahead. There were some significant changes to the cmake setup of the project recently (from #562), so it might require a significant update to the code for the new variant. Would you like me to handle the rebase on a separate PR for you (crediting the changes to you as co-author)? Or would you prefer to do it yourself?

I'm happy to let you get it fixed up and ready to merge.

PR #527, opened 2024-10-07
Revert "Adjust clang's multilib flags test to cope with missing variant"

pratlucas:
Reverts ARM-software/LLVM-embedded-toolchain-for-Arm#504.

The upstream change from llvm/llvm-project#109640 has now been merged, allowing tests to specify multilib yaml file using a new command line option. The relevant tests have been updated by that change, so the downstream adjustment is no longer needed.


ostannard (approved):
LGTM

PR #529, opened 2024-10-10
[CMake] Fix default install directory

VladiKrapp-Arm:
The mechanism for setting the default installation directory to <BUILDDIR>/install is highly dependent on the order of events in the CMakeLists.txt file. Due to recent shifts in the file this has stopped working.
The CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT check must happen immediately after the first invocation of project(...). When the llvm-project directory is included first, it is calling project() and then sets the CMAKE_INSTALL_PREFIX to the system default.


ostannard (approved):
LGTM

PR #531, opened 2024-10-11
Set output names for libcxx shared libraries

### LINKED ISSUE #530: Problem with upstream commit 917ada35
Hello
In order to build the toolchain from the latest commits of upstream I need to use the following workaround:
Revert "[runtimes] Always define cxx_shared, cxx_static & other targets (#80007)"
(upstream commit 917ada35).
I presume that is needed in LLVM-ET-for-Arm CmakeLists.txt some lines to handle variables:
- LIBCXX_ENABLE_SHARED
- LIBUNWIND_ENABLE_SHARED

dcandler:
Although building shared libraries is not enabled, CMake still creates targets for them, and implicitly changes them to static. In order to prevent this duplication of library targets, this patch explicitly sets the output names for shared libraries to something different.

Closes #530


ostannard (approved):
LGTM, this matches the fix suggested in https://github.com/llvm/llvm-project/pull/80007#issuecomment-2402188675.

PR #532, opened 2024-10-14
Fix library variants for v7-R soft-float ABI

ostannard:
We currently build these two library variants for v7-R with the soft-float ABI:
* armv7r_soft_nofp
* armv7r_soft_vfpv3_d16

The first variant can only be used when the target does not have an FPU, because it does not include bootcode to enable the FPU. This means that we don't have any library variants which can be selected for the single-precision-only FPUs when using the soft-float ABI. For the hard-float ABI, we solve this by having separate library variants for vfpv3xd (SP-only) and vfpv3_d16 (SP+DP). This is important for performance, as it allows the math library to use DP instructions where available.

We could do the same thing for the soft-float libraries, however these variants with an FPU are included for completeness, their performance is less important because a user interested in performance would use the hard-float ABI. Thesefore I think we should only include the variant for the minimal FPU, which is vfpv3xd.

These libraries are also used for v8-R AArch32, where this change is important because -march=armv8-r implies -mfpu=fpv5-sp-d16 by default, which currently fails to find a library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1768,9 +1768,9 @@ add_library_variants_for_cpu(
 )
 add_library_variants_for_cpu(
     armv7r
-    SUFFIX soft_vfpv3_d16
-    COMPILE_FLAGS "-mfloat-abi=softfp -march=armv7r -mfpu=vfpv3-d16"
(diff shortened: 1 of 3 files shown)


simpal01:
LGTM

simpal01 (approved):

PR #533, opened 2024-10-14
Add -enable-dfa-jump-thread to Omax.cfg

stuij:
This optimization sees some good improvements and the downstream equivalent is enabled for -Omax on Arm Compiler 6.


ostannard (approved):
LGTM

PR #536, opened 2024-10-15
[Pipelines] Additional unrolling  in LTO

VladiKrapp-Arm:
Some workloads require specific sequences of events to happen
to fully simplify. This adds an extra full unrolling pass to help these
cases on the cores with branch predictors. It helps produce simplified
loops, which can then be SROA'd allowing further simplification, which
can be important for performance.
Feature adds extra compile time to get extra performance and 
is enabled by the opt flag 'extra-LTO-loop-unroll' (off by default).

Original patch by David Green (david.green@arm.com)


patches/llvm-project-perf/0000-LTOpasses-add-loop-unroll.patch
```diff
@@ -0,0 +1,55 @@
...
+
+Some workloads require specific sequences of events to happen
+to fully simplify. This adds an extra full unrolling pass to help these
+cases on the cores with branch predictors. It helps produce simplified
+loops, which can then be SROA'd allowing further simplification, which
+can be important for performance.
+This is added under own flag - spending extra compile time to get extra
+performance on specific user request.
+
+Originally patch by David Green (david.green@arm.com)
```
stuij, line 14: Originally -> Original

patches/llvm-project-perf/0000-LTOpasses-add-loop-unroll.patch
```diff
@@ -0,0 +1,55 @@
...
+From: Vladi Krapp <vladi.krapp@arm.com>
+Date: Tue, 3 Sep 2024 14:12:48 +0100
+Subject: [Pipelines] Additional unrolling  in LTO 
+
+Some workloads require specific sequences of events to happen
+to fully simplify. This adds an extra full unrolling pass to help these
+cases on the cores with branch predictors. It helps produce simplified
+loops, which can then be SROA'd allowing further simplification, which
+can be important for performance.
+This is added under own flag - spending extra compile time to get extra
```
stuij, line 11: I would name the flag here in the description, so people don't have to hunt through the commit itself to find it.

under own flag -> under it's own flag

these comments also apply to the description of this patch itself, which has the same text

patches/llvm-project-perf/0000-LTOpasses-add-loop-unroll.patch
```diff
@@ -0,0 +1,55 @@
...
+performance on specific user request.
```
stuij, line 12: `on specific user request` isn't quite correct English. I would just remove it.

dcandler:
Very minor nit: the first patch should start `0001`, i.e. `0001-LTOpasses-add-loop-unroll.patch` to match the other folder and the likely output of `git format-patch`
The placeholder patch was numbered 0 since it's not a real patch file.

VladiKrapp-Arm:
@stuij , @dcandler , I have made the suggested changes. 
Any other issues?

PR #538, opened 2024-10-16
Add LLVM libc sample

smithp35:
This will be built into the LLVM libc overlay package. Sample installs to samples/src/llvmlibc/baremetal-semihosting.

* Add llvmlibc-samples directory to contain LLVM libc specific samples.
* Add single baremetal-semihosting sample using LLVM libc.

Sample contains a number of calls to LLVM libc functionality:
* heap.
* printf.
* string library.
* math library.

A special linker script is used to set the bounds of the heap. This uses different linker defined symbols to picolibc.

The crt0.S from the existing llvmlibc-support can't be used directly as the RW and ZI from the linker script need initializing.

### DIFF
--- a/llvmlibc-samples/src/llvmlibc/baremetal-semihosting/Makefile
+++ b/llvmlibc-samples/src/llvmlibc/baremetal-semihosting/Makefile
@@ -1,0 +1,37 @@
+#
+# Copyright (c) 2020-2023, Arm Limited and affiliates.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+include ../../../Makefile.conf
+
+build: hello.elf
+
+hello.elf: *.c
+	../$(BIN_PATH)/clang --config=llvmlibc.cfg $(MICROBIT_TARGET) -nostartfiles -lsemihost -g -fno-exceptions -fno-rtti -T microbit-llvmlibc.ld -o hello.elf $^
+
+%.hex: %.elf
+	../$(BIN_PATH)/llvm-objcopy -O ihex $< $@
+
+run: hello.hex
+	qemu-system-arm -M microbit -semihosting -nographic -device loader,file=$<
+
+debug: hello.hex
+	qemu-system-arm -M microbit -semihosting -nographic -device loader,file=$< -s -S
+
+clean:
+	rm -f *.elf *.hex
+
+.PHONY: clean run debug
--- a/llvmlibc-samples/src/llvmlibc/baremetal-semihosting/Makefile
+++ b/llvmlibc-samples/src/llvmlibc/baremetal-semihosting/Makefile
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2020-2023, Arm Limited and affiliates.
+# Copyright (c) 2020-2024, Arm Limited and affiliates.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/llvmlibc-samples/src/llvmlibc/baremetal-semihosting/crt0llvmlibc.c
+++ b/llvmlibc-samples/src/llvmlibc/baremetal-semihosting/crt0llvmlibc.c
@@ -34,8 +34,8 @@ extern char __tdata_end[];
 extern char __tls_end[];
 
 void _start(void) {
-  memcpy(__data_start, __data_source, (uintptr_t) __data_size);
-  memset(__bss_start, '\0', (uintptr_t) __bss_size);
+  memcpy(__data_start, __data_source, (size_t) __data_size);
+  memset(__bss_start, 0, (size_t) __bss_size);
   _platform_init();
   _Exit(main(0, NULL));
 }
--- a/llvmlibc-samples/src/llvmlibc/baremetal-semihosting/crt0llvmlibc.c
+++ b/llvmlibc-samples/src/llvmlibc/baremetal-semihosting/crt0llvmlibc.c
@@ -1,0 +1,41 @@
+// Copyright (c) 2024, Arm Limited and affiliates.
+// SPDX-License-Identifier: Apache-2.0
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+
+//    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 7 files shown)


llvmlibc-samples/src/llvmlibc/baremetal-semihosting/microbit-llvmlibc.ld
```diff
@@ -0,0 +1,240 @@
+/*
+ * SPDX-License-Identifier: BSD-3-Clause
+ *
+ * Copyright © 2019 Keith Packard
```
voltur01, line 4: Do we need to add Arm copyright as well?

pratlucas:
What's the motivation for using a separate directory structure instead of placing the new samples inside the existing `samples` directory?

llvmlibc-samples/src/llvmlibc/baremetal-semihosting/microbit-llvmlibc.ld
smithp35: It was a trivial modification, but it is probably best to add the additional copyright anyway.

llvmlibc-samples/src/llvmlibc/baremetal-semihosting/Makefile
```diff
@@ -0,0 +1,37 @@
+#
+# Copyright (c) 2020-2023, Arm Limited and affiliates.
```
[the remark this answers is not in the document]
smithp35: OK, I'll add this change in. Given I'm making other changes I'll update in a follow up commit.

llvmlibc-samples/src/llvmlibc/baremetal-semihosting/crt0llvmlibc.c
```diff
@@ -0,0 +1,41 @@
...
+extern char __data_size[];
+extern char __bss_start[];
+extern char __bss_end[];
+extern char __bss_size[];
+extern char __tls_base[];
+extern char __tdata_end[];
+extern char __tls_end[];
+
+void _start(void) {
+  memcpy(__data_start, __data_source, (uintptr_t) __data_size);
```
statham-arm, line 37: Pedantry: formally, memcpy and memset take a `size_t`, not a `uintptr_t`.

llvmlibc-samples/src/llvmlibc/baremetal-semihosting/hello.c
```diff
@@ -0,0 +1,38 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <math.h>
+
+/* Example that uses heap, string and math library */
```
statham-arm, line 22: I'm not sure you have used the math library, have you? I can only see a call to `abs`, which is defined in `stdlib.h`, not `math.h`.

The problem with the actual math.h functions in embedded LLVM libc is that they won't make a good demo, because you can't printf the floating-point results, because FP printf is left out in the embedded build profile!

I suppose if you really wanted to demonstrate math.h functions, you could do something that has an integer final result, along the lines of `lround(400000 * atan(1))` to get 314159. But it's probably also worth adding a comment explaining _why_ you picked that example, so that when the user inevitably tries tweaking it to printf a floating-point value, they're warned why it won't work.

llvmlibc-samples/src/llvmlibc/baremetal-semihosting/crt0llvmlibc.c
smithp35: Thanks will fix.

llvmlibc-samples/src/llvmlibc/baremetal-semihosting/hello.c
smithp35: That will teach me for using the list of math functions in cppreference. Was trying to find one that returned an integer.

I've used `atanf` as llvm-libc doesn't define `atan`. This also showed up that I hadn't defined `int *__llvm_libc_errno()` so I've added that to the docs aswell.

smithp35:
> > > What's the motivation for using a separate directory structure instead of placing the new samples inside the existing `samples` directory?
> > 
> > 
> > A couple of reasons, neither particularly strong:
> > 
> > * Easier to avoid name clashing with the existing picolibc samples as the overlay package is developed separately. This could be resolved by explicitly adding llvmlibc to each sample name, for example baremetal-semihosting-llvmlibc.
> > * It made it easier for me to remove the llvm-libc samples when repeatedly testing the package, but I think that's resolvable.
> > 
> > I'm open to opinions on what's best?
> 
> For consistency I'd suggest keeping everything under `samples`. It also helps to keep the directory structure simple and clear.
> 
> Could we maybe have separate sub-sirectories inside `samples` for each library package we provide? E.g.:
> 
> ```
> LLVM-embedded-toolchain-for-Arm/
> ├─ samples/
> │  ├─ picolibc/
> │  ├─ newlib/
> │  ├─ llvmlibc/
> ├─ ...
> ```

Are you OK with me doing that in a follow up change as this one currently only affects the overlay package?

llvmlibc-samples/src/llvmlibc/baremetal-semihosting/hello.c
```diff
@@ -19,7 +19,13 @@
 #include <string.h>
 #include <math.h>
 
-/* Example that uses heap, string and math library */
+// Implementation of errno
+int *__llvm_libc_errno() {
+  static int errno;
```
statham-arm, line 24: Sorry to do another pedantic one, but I recommend calling that internal variable something other than `errno`. If a user starts from this code and adds `#include <errno.h>`, then the name `errno` will surely become a macro, making this function into nonsense. Any other name is fine.

smithp35: Thanks for spotting. I've used a name that doesn't contain errno.

llvmlibc-samples/src/llvmlibc/baremetal-semihosting/hello.c
```diff
@@ -0,0 +1,46 @@
...
+  const size_t world_s_len = strlen(world_s);
+  const size_t out_s_len = hello_s_len + world_s_len + 1;
+  char *out_s = (char*) malloc(out_s_len);
+  assert(out_s_len >= hello_s_len + 1);
+  strncpy(out_s, hello_s, hello_s_len + 1);
+  assert(out_s_len >= strlen(out_s) + world_s_len + 1);
+  strncat(out_s, world_s, world_s_len + 1);
+  // 2024-10-17 Embedded printf implementation does not currently
+  // support printing floating point numbers.
+  printf("%s %li\n", out_s, lround(400000 * atanf(1.0f)));
```
statham-arm, line 43: Because you wrote `lround` rather than `lroundf`, the float returned from `atanf` is being converted into a double, so this code is testing a mix of single and double precision operations.

I think that's actually a good thing – it makes it a better test! I point it out because you probably didn't do it on purpose, but I quite like it, and perhaps it's better not to "fix" it :-)

statham-arm (approved):
Thanks for the fixes. I'm happy with this now, but of course other reviewers have been commenting on entirely different aspects of the patch.

voltur01 (approved):
LGTM too, thank you!

pratlucas (approved):
> Are you OK with me doing that in a follow up change as this one currently only affects the overlay package?

Yes, that's ok. LGTM.

PR #540, opened 2024-10-23
Remove armv4t/armv5t from check-all targets

dcandler:
The compiler-rt and libcxx tests for armv4t and armv5t variants currently hang until a timeout is reached. Rather than disable the tests, this patch removes them from the check-all targets. This allows the tests to still be run while investigating the issue, but does not interrupt overall testing of the toolchain.


dcandler:
Minor update to also include picolibc tests, which are also timing out (albeit with a more sensible limit)

PR #543, opened 2024-10-26
Fix MinGW toolchain install with 'install-llvm-toolchain'

stephanosio:
When MinGW LLVM Embedded Toolchain for Arm is installed using the `install-llvm-toolchain` target, the MinGW-built toolchain executables are not installed under the prefix because `llvm-toolchain-mingw` component is added to the `LLVM_TOOLCHAIN_DISTRIBUTION_COMPONENTS_ALL` variable, which is never referenced.

Fix this by enumerating `LLVM_TOOLCHAIN_DISTRIBUTION_COMPONENTS_ALL` as intended.


voltur01:
This would be the alternative implementation https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/pull/613 that I cannot test right now since mingw GCC on my Linux is too old to build current LLVM version.

PR #544, opened 2024-10-28
[OmaxLTO] Fix the flag to be supplied to the LTO plugin

VladiKrapp-Arm:
The flag to add the extra loop unroll pass was previously provided to opt and was not actually working, as it's supposed to be provided to the LTO plugin.


SamaliLiyanage (approved):
LGTM

PR #545, opened 2024-10-28
Use off-the-shelf ccache enablement

vhscampos:
LLVM's CMake system offers an off-the-shelf detection and enablement of ccache in its builds: LLVM_CCACHE_BUILD.

This patch replaces our manually written mechanism by the one that LLVM offers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,14 +223,7 @@ endif()
 
 find_package(Python3 REQUIRED COMPONENTS Interpreter)
 
-if(NOT CMAKE_C_COMPILER_LAUNCHER AND NOT CMAKE_CXX_COMPILER_LAUNCHER)
-    # If ccache is available then use it by default.
-    find_program(CCACHE_EXECUTABLE ccache)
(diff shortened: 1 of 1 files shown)


ostannard (approved):
LGTM

statham-arm:
What's the effect of this on subsidiary builds like libc and libc++? I guess _neither_ method would pass through a request to use ccache, so this doesn't make anything worse?

statham-arm (approved):
... yes, discussing this elsewhere, we think that ccache was not used for library builds, before _or_ after this change. So I'm happy that this is an unconditional improvement.

(@vhscampos also points out that ccache may do the wrong thing if the compiler has changed, so that if you're rebuilding the whole toolchain, you don't _want_ to ccache the library builds. So that might have been a deliberate choice, although it doesn't suit everybody -- library developers who aren't constantly changing the compiler would probably prefer more ccache.)

vhscampos:
Thanks. Since `LLVM_CCACHE_BUILD` is in the path of deprecation, there's no point in this PR.

PR #546, opened 2024-10-28
Fix Python SyntaxWarning in two f-string literals.

statham-arm:
Building on Ubuntu 24.04, multilib-generate.py provokes some SyntaxWarning messages from Python 3.12, because of cases where we'd written `\+` in a string, intended to be emitted literally. 3.12 still _does_ emit it literally, but prints a warning pointing out that we should properly escape the backslash, writing it as `\\+`.

### DIFF
--- a/multilib-generate.py
+++ b/multilib-generate.py
@@ -247,10 +247,10 @@ def generate_extensions(args):
     print("# original option to check that.")
 
     for feature in all_features:
-        print(f"- Match: -march=armv.*\\+{feature}($|\+.*)")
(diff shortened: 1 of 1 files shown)


voltur01 (approved):

ostannard (approved):
LGTM. Github made this confusing by removing some of the backslashes in the PR summary, you'll probably need to re-add them before merging.

statham-arm:
Or rather, use backticks so that it will look sensible whether you read the commit message as Markdown _or_ as plain text. I don't want to _force_ people to look at commits on the website!

PR #547, opened 2024-10-31
Add aarch64_sme_accessible to newlib patch

Rin18:
To check for SME support, compiler-rt makes a call to __aarch64_sme_accessible(). It expects this symbol to be defined in newlib, which is currently incorrect. This patch extends newlib.patch to add the definition of __aarch64_sme_accessible.

### DIFF
--- a/patches/newlib.patch
+++ b/patches/newlib.patch
@@ -1,3 +1,20 @@
+diff --git a/libgloss/aarch64/syscalls.c b/libgloss/aarch64/syscalls.c
+index 7343cc61f..2c4b63c17 100644
+--- a/libgloss/aarch64/syscalls.c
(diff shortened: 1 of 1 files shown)


statham-arm (commented):
Could you add some more explanation in the commit message, please?

Who is expecting to call this function? What currently isn't working because it's missing?

Does it have a specification anywhere? Or does another C library implement it, that you're aiming to match the API of?

Rin18:
> Could you add some more explanation in the commit message, please?
> 
> Who is expecting to call this function? What currently isn't working because it's missing?
> 
> Does it have a specification anywhere? Or does another C library implement it, that you're aiming to match the API of?

I've changed the commit message to add more explanation.

statham-arm (approved):

PR #550, opened 2024-10-31
Add combinations of commands for which multi-lib is found successfully

vrukesh:
Add combinations of commands for which multi-lib is found successfully


test/multilib/multilib_success_commands.test
```diff
@@ -0,0 +1,1936 @@
+; Success
+
```
pratlucas, line 2: Can this test be split into multiple files? It'd help with maintainability and allow things to be tested in parallel.
A suggestion for the initial split would be something like:
- aarch64_a_supported_variants.test
- aarch64_r_supported_variants.test
- arm_a_supported_variants.test
- arm_r_supported_variants.test
- arm_m_supported_variants.test

pratlucas (approved):
LGTM.

pratlucas (approved):
LGTM.

PR #551, opened 2024-11-01
[libcxx] Remove xfails for char16_t

vhscampos:
picolibc added support for `char16_t` on trunk. However we can't yet remove the xfails in upstream because it firstly needs to pick up on a picolibc release that includes that support.

This is a patch which must be upstreamed once llvm-project updates its picolibc version to one that has `char16_t` support.


pratlucas (approved):
LGTM.

PR #553, opened 2024-11-04
Add Armv7a NO FP -mno-unaligned-access library variants

simpal01:


ostannard:
> `strictly_aligned`

I think it would be better to keep this as one word with no underscores, so that it is clearer how these parse. `exn_rtti` doesn't follow that, but it could be considered two separate features, and a `rtti` variant without exceptions could exist in future (not that I think it _should_).

Since we are eventually going to want strict-aligned variants for most architectures, would it make more sense to handle it like we do for `exn_rtti` and expand the library variants inside `add_library_variants_for_cpu`, instead of adding more calls to it? That will need an extra parameter for now to only enable the new variants for architectures where they work, but will avoid a lot of duplication in the long run, especially when we add big-endian variants which would double everything again if done in the same way.

pratlucas (approved):
LGTM.

PR #554, opened 2024-11-04
Add AArch64 Big Endian Variants using FVP models

simpal01:
This patch add new aarch64 big endian library variants.
 1. aarch64_be
 2. aarch64_be_exn_rtti

QEMU's doesn't have big-endian support in the release binaries. so need to use FVPs to test those libraries.


simpal01:
> Similar comment to #553: should we move the expansion of library variants inside the `add_library_variants_for_cpu` function, like we do for exceptions/rtti,instead of adding more calls to it? That will probably make this patch a bit more complicated, as we'll need to pass both the QEMU and FVP parameters into the call and select the right one for each variant, but in the longer term we'll want big-endian and strict-align variants for (almost) every library variant, so it should simplify things in the end.

Yeah there is a plan to do this as a long term solution when you have more library variants work with big-endian and strict-align. As a starting point, we will start adding separately now.

arm-multilib/json/variants/aarch64a_be.json
```diff
@@ -0,0 +1,40 @@
+{
+    "args": {
+        "common": {
+            "TARGET_ARCH": "aarch64a",
+            "VARIANT": "aarch64a_be",
+            "COMPILE_FLAGS": "-march=armv8-a -mbig-endian",
+            "ENABLE_EXCEPTIONS": "OFF",
+            "ENABLE_RTTI": "OFF",
+	    "TEST_EXECUTOR": "fvp",
```
dcandler, line 9: Tab has crept in instead of whitespace.

arm-multilib/json/variants/aarch64a_be_exn_rtti.json
```diff
@@ -0,0 +1,40 @@
+{
+    "args": {
+        "common": {
+            "TARGET_ARCH": "aarch64a",
+            "VARIANT": "aarch64a_be_exn_rtti",
+            "COMPILE_FLAGS": "-march=armv8-a -mbig-endian",
+            "ENABLE_EXCEPTIONS": "ON",
+            "ENABLE_RTTI": "ON",
+	    "TEST_EXECUTOR": "fvp",
```
dcandler, line 9: Ditto

dcandler (approved):

PR #562, opened 2024-11-11
Move CMake code for building multilib and variants to subprojects

dcandler:
This patch restructures the top-level CMake script in order to split building the library variants into a separate CMake project. This better encapsulates the code to build the variants, and untethers it from the toolchain builds so that libraries can be developed and tested without having the rebuild the entire toolchain.

Similarly, a new subproject has been created to build a multilib set of library variants, using a selected C library. A simple JSON format has been used to define the flags and variants, which are then built and collected into the appropriate layout.


arm-multilib/CMakeLists.txt
```diff
@@ -0,0 +1,243 @@
...
+    list(APPEND compiler_launcher_cmake_args "-DCMAKE_C_COMPILER_LAUNCHER=${CMAKE_C_COMPILER_LAUNCHER}")
+endif()
+if(CMAKE_CXX_COMPILER_LAUNCHER)
+    list(APPEND compiler_launcher_cmake_args "-DCMAKE_CXX_COMPILER_LAUNCHER=${CMAKE_CXX_COMPILER_LAUNCHER}")
+endif()
+
+# Arguments to pass down to the library projects.
+foreach(arg
+    LLVM_BINARY_DIR
+    FVP_INSTALL_DIR
```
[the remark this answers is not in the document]
dcandler: Yes, these (and some other options such as `LBC_HDRGEN`) were missed and will be added in a fixup.

arm-multilib/multilib-generate.py
```diff
@@ -238,7 +238,8 @@ def get_extension_list(clang, triple):
 def generate_extensions(args):
     aarch64_features = get_extension_list(args.clang, "aarch64-none-eabi")
     aarch32_features = get_extension_list(args.clang, "arm-none-eabi")
-    all_features = set(aarch64_features) | set(aarch32_features)
+    all_features = list(aarch64_features)
+    all_features.extend(feat for feat in list(aarch32_features) if feat not in all_features)
```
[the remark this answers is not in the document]
dcandler: Just determinism: I was trying to check the generated `multilib.yaml` file was not any different after my changes in this patch, but discovered the output of this script is different every time it is run because of the casting to sets. Otherwise order of the features doesn't seem to be important; I might have thought it neater to have aarch64 features grouped together though a sort would be easier.

arm-runtimes/CMakeLists.txt
```diff
@@ -0,0 +1,780 @@
...
+            lit_test_executor
+            ${CMAKE_CURRENT_SOURCE_DIR}/test-support/lit-exec-fvp.py
+            ${test_executor_params}
+        )
+    endif()
+    list(JOIN lit_test_executor " " lit_test_executor)
+endif()
+
+set(compile_arch_flags "--target=${target_triple} ${COMPILE_FLAGS}")
+set(link_flags "-nostartfiles -lcrt0-semihost -lsemihost -T picolibcpp.ld")
```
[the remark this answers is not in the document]
dcandler: Probably would be better named `test_link_flags` since it only gets used for linking compiler-rt and libcxx tests. It would sit better if moved slightly above into the `if` block that is only enabled with tests being enabled.
At the moment testing only works with picolibc, so a lot of the existing code setting up the tests was only written with picolibc in mind and will need changing for other libraries. For now attempting to enable tests will simply result in a configuration error.

arm-runtimes/test-support/modify-compiler-rt-xml.py
```diff
@@ -0,0 +1,54 @@
...
+
+import argparse
+import os
+from xml.etree import ElementTree
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Reformat compiler-rt xml results")
+    parser.add_argument(
+        "--compiler-rt-build-dir",
```
statham-arm, line 17: Small nit: I'm not a great fan of enormously long option names in general, but especially not if they're mandatory. If I _have_ to specify this option, can't we at least have a `-d` alias for it or something?

docs/building-from-source.md
```diff
@@ -157,3 +157,83 @@ and designed to improve performance in certain circumstances.
...
+`ninja check-compiler-rt`
+`ninja check-cxx`
+`ninja check-cxxabi`
+`ninja check-unwind`
+
+Alternatively, `ninja check-all` runs all enabled tests. 
+
+## Building sets of libraries
+
+As well as indvidual libraries, it is also possible to build a set of
```
statham-arm, line 203: tyop: "individual"

arm-multilib/CMakeLists.txt
```diff
@@ -0,0 +1,243 @@
...
+                set(parent_dir_name aarch64-none-elf)
+            else()
+                set(parent_dir_name arm-none-eabi)
+            endif()
+            set(destination_directory "${CMAKE_CURRENT_BINARY_DIR}/multilib/${parent_dir_name}/${variant}")
+            install(
+                DIRECTORY ${destination_directory}
+                DESTINATION ${parent_dir_name}
+            )
+            set(variant_json_file ${CMAKE_CURRENT_BINARY_DIR}/json/variants/${variant_json})
```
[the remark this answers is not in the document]
statham-arm: OK – so we might later add something like a `compatible_libcs` field in the JSON and filter on that? That seems reasonable.

arm-runtimes/CMakeLists.txt
```diff
@@ -0,0 +1,795 @@
...
+endif()
+
+# Create a single target for all testing. If no testing is enabled, this
+# will simply do nothing.
+add_custom_target(check-all)
+
+# If any testing is enabled, prepare test executor settings.
+if(ENABLE_LIBC_TESTS OR ENABLE_COMPILER_RT_TESTS OR ENABLE_LIBCXX_TESTS)
+    # Flags required to link tests.
+    set(test_link_flags "-nostartfiles -lcrt0-semihost -lsemihost -T picolibcpp.ld")
```
statham-arm, line 156: I'm still confused about these flags being picolibc-specific. One of the conditions that enables this if statement is `ENABLE_LIBC_TESTS`. But that variable is documented further up this file as `"Enable libc tests (picolibc, newlib or llvm-libc)."`

Either we can run tests on other libcs or we can't, surely? If we can't then there should be a check here that throws an error; if we can then this should be an if statement that sets up appropriate flags?

dcandler: There are checks to throw an error for attempting to enable testing with anything other than picolibc selected, but it's done later in the library specific blocks.

But you are right in that it should be clear that these will also need updating when we start adding more testing. Additionally throwing an error here is probably the safest way to ensure the currently hardcoded flags don't get missed.

arm-runtimes/test-support/modify-compiler-rt-xml.py
dcandler: I think this was a holdover from an earlier version that required multiple directories. There's no reason for the length now.

statham-arm (approved):
LGTM this time, thanks!

PR #563, opened 2024-11-13
[Perf] Update inlining memcpy patch

nasherm:
The patch file for setting inilining memcpy preference was slightly incorrect. Use of the SmallVector data structure caused issues with correctly inlining on some tests. The optimisation was also not enabled for v8.1m mainline targets as it should be. I've updated the patch file to reflect this.


VladiKrapp-Arm (approved):
LGTM

You could also document the flag in docs/optimization-flags.md

PR #564, opened 2024-11-13
Downstream patch to add bootcode for AArch64 FVPs #852

simpal01:
The AArch64 FVP (Fixed Virtual Platform) models differ from QEMU in a
    few ways which affect the crt0 code:
   
    * The memory map is different, so needs different page tables.
    * They boot up at EL3, instead of EL1, so we need to set the EL3 versions of the system registers.
    * Add option to build crt0 bootcode for different machines
    * Build new FVP variants of crt0, instead of replacing QEMU ones
    * Split assembly parts of AArch64 crt0 into a separate file
    * Make error checking target-specific
    
Authored by: **Oliver Stannard**

### DIFF
--- a/patches/picolibc.patch
+++ b/patches/picolibc.patch
@@ -35,3 +35,850 @@ index 7b63ba172..cda5e1e7e 100644
  		*(.gnu.linkonce.t.*)
  		KEEP (*(.fini .fini.*))
  		@PREFIX@__text_end = .;
+diff --git a/meson.build b/meson.build
+index 8cf1006d9..4e612966f 100644
+--- a/meson.build
++++ b/meson.build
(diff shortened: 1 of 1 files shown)


dcandler:
If we need multiple changes to picolibc, would it help to create a subfolder and split them into distinct patch files like we did for `llvm-project`? So that it is easier to track what is being changed and why, instead of having everything mixed into a single diff file.

simpal01:
> If we need multiple changes to picolibc, would it help to create a subfolder and split them into distinct patch files like we did for `llvm-project`? So that it is easier to track what is being changed and why, instead of having everything mixed into a single diff file.

Done

dcandler (approved):

PR #567, opened 2024-11-15
Add aarch64 library variants for no-unaligned-access

pratlucas:
This introduces new library variants for aarch64 to support targets
where unaligned memory accesses are disabled.
To enable these changes a patch to picolibc was required, ensuring the
assembly implemenation that relies on unaligned accesses is not used
when `__ARM_FEATURE_UNALIGNED` is not defined.


simpal01 (approved):

pratlucas:
Abandoning this PR in favour of #572.

PR #571, opened 2024-11-18
Corrections to CMake changes to enable library sub-builds

dcandler:
Two scripts arguments were renamed during review, but the cmake steps calling the script weren't updated.

Additionally the our downstream test targets such as check-package-llvm-toolchain target require the path to an external llvm-lit in order to run the tests.


DaftanoPro (approved):
look goods to me

dcandler:
Found one more issue: the parameter to enable exceptions and RTTI was not getting passed down correctly so the exceptions sample failed.

PR #573, opened 2024-11-21
CMakeLists.txt: tiny fix to instructions comment.

statham-arm:
Now that our downstream picolibc changes are in the form of multiple patches in a subdirectory, it's better to apply them with 'git am' so they turn into separate commits, instead of 'git apply', the same way we do it for llvm-project.


statham-arm:
Good catch, thanks!

simpal01 (approved):

PR #575, opened 2024-11-21
Docs fixes for arm-runtimes and arm-multilib sub-builds.

statham-arm:
The suggested build commands left out the usual `cd` after `mkdir`, so that if you followed them literally, you'd create an `arm-runtimes` or `arm-multilib` directory and then ignore it completely and put all your build debris at the level above it.

Also, trying to follow the instructions, I was confused by the semantics of `LLVM_BINARY_DIR`: given the name, I instinctively pointed it at the actual `bin` directory, but it turned out I should have aimed one level up. Added some clarifying text, with particular reference to using a full build of this toolchain for a standalone build of one library.

While I'm here, removed some apparently accidental physical tabs, so that all the continuation lines in the long cmake commands line up.

### DIFF
--- a/docs/building-from-source.md
+++ b/docs/building-from-source.md
@@ -171,7 +171,13 @@ format in the arm-multilib/json/variants folder, which can be loaded at
 configuration with the `-DVARIANT_JSON` setting. Any additional options
 provided on the command line will override values from he JSON. `-DC_LIBRARY`
 will be required to set which library to build, and `-DLLVM_BINARY_DIR` should
-point to a build or install of LLVM.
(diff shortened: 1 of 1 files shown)


dcandler (approved):
Thanks - I can understand the confusion around `LLVM_BINARY_DIR` since it's using "binary" in the CMake context of the build tree rather than a source one. But calling it the build directory also didn't quite work since it can also be pointed to an install or extracted package.

statham-arm:
Yes, I guessed that was the sense you meant "binary" in. As you say – it's a valid use, just not the only use, so the clarification seems worth having.

PR #578, opened 2024-11-26
Rename the variant name  from strictly_aligned to strictalign.

simpal01:
This is to match with the corresponding aarch64 variant for -mno-unaligned-access.


pratlucas (approved):
LGTM.

PR #586, opened 2024-12-02
Add an option to disable QEMU testing

dcandler:
Testing with QEMU is implicitly enabled by default, but configuring the project to run with tests requires QEMU. As testing with QEMU is considered optional, it should be possible to disable tests when QEMU is not available or testing is not needed.

Since there is already an option for controlling FVP testing, this patch adds a similar option for QEMU, ENABLE_QEMU_TESTING, which instead defaults to ON. This can then be turned off when QEMU is not installed, so that configurations where QEMU is expected can suitably detect and show an error when it is missing.

### DIFF
--- a/arm-multilib/CMakeLists.txt
+++ b/arm-multilib/CMakeLists.txt
@@ -34,6 +34,11 @@ set_property(CACHE C_LIBRARY PROPERTY STRINGS picolibc newlib llvmlibc)
 set(LLVM_BINARY_DIR "" CACHE PATH "Path to LLVM toolchain build or install root.")
 set(LIBC_HDRGEN "" CACHE PATH "Path to prebuilt lbc-hdrgen if not included in LLVM binaries set by LLVM_BINARY_DIR")
 option(
+    ENABLE_QEMU_TESTING
+    "Tests using QEMU are enabled by default, but can be disabled."
+    ON
+)
+option(
     ENABLE_FVP_TESTING
     "Tests using FVP need to be explictly enabled."
 )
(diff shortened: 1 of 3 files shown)


arm-multilib/CMakeLists.txt
```diff
@@ -163,7 +168,7 @@ foreach(lib_idx RANGE ${lib_count_dec})
 
             # FVP testing should default to off, so override any
```
smithp35, line 169: Does this comment need to be revised now that both FVP and QEMU testing is optional?

arm-multilib/CMakeLists.txt
```diff
@@ -33,6 +33,11 @@ set(C_LIBRARY "picolibc" CACHE STRING "Which C library to use.")
 set_property(CACHE C_LIBRARY PROPERTY STRINGS picolibc newlib llvmlibc)
 set(LLVM_BINARY_DIR "" CACHE PATH "Path to LLVM toolchain build or install root.")
 set(LIBC_HDRGEN "" CACHE PATH "Path to prebuilt lbc-hdrgen if not included in LLVM binaries set by LLVM_BINARY_DIR")
+option(
+    ENABLE_QEMU_TESTING
+    "Tests using QEMU are enabled by default, but can be disabled."
```
smithp35, line 38: The help text, while correct, doesn't fit well with `ENABLE_QEMU_TESTING`. It would fit better with a DISABLE_QEMU_TESTING option. 

Perhaps
```
Enable tests that use QEMU. This option is ON by default.
```

If the text is changed, would be good to update the other places too.

smithp35 (commented):
Code change LGTM, I think we may want to change the help text.

arm-multilib/CMakeLists.txt
dcandler: Done, and updated FVP options to match.

smithp35 (approved):
LGTM, thanks for the update.

PR #587, opened 2024-12-02
Prevent arm-runtime CMake arguments persisting over loops

dcandler:
The additional_cmake_args variable is used to override the defaults and disable testing so that it can be controlled by the higher-level CMake project. But the variable is currently set only when the tests should be disabled. If the tests do not need to be disabled, the value is not updated and the version from the previous iteration of the loop will be used, potentially incorrectly disabling the tests.

To fix this, clear the value on every loop if not needed.

### DIFF
--- a/arm-multilib/CMakeLists.txt
+++ b/arm-multilib/CMakeLists.txt
@@ -136,6 +136,7 @@ math(EXPR lib_count_dec "${lib_count} - 1")
 foreach(lib_idx RANGE ${lib_count_dec})
     string(JSON lib_def GET ${multilib_defs} ${lib_idx})
     string(JSON variant GET ${lib_def} "variant")
+    set(additional_cmake_args "")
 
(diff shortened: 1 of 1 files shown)


arm-multilib/CMakeLists.txt
```diff
@@ -169,6 +169,7 @@ foreach(lib_idx RANGE ${lib_count_dec})
                 set(read_ENABLE_COMPILER_RT_TESTS "OFF")
                 set(read_ENABLE_LIBCXX_TESTS "OFF")
             else()
+                set(additional_cmake_args "")
```
smithp35, line 172: Would this be better off above the if statement? That way if another conditional is added then it will automatically pick up a default "".

dcandler: Good point - I've changed it so every loop starts with an empty list, and the args get appended which makes more sense.

smithp35 (approved):
LGTM. Thanks for the update.

PR #588, opened 2024-12-04
Add aarch64a_soft_nofp variant

vhscampos:
Build a soft float multilib variant for aarch64 supporting targets without an FPU.

Authored by Keith Packard and Victor Campos.


vhscampos:
CC @keith-packard

pratlucas (approved):
LGTM

PR #589, opened 2024-12-06
Changes to AArch64 nofp variant

vhscampos:
- Make the variants strictly aligned
- Create new variant for big endian
- Create new variant for exceptions and RTTI


smithp35 (approved):
LGTM. I guess that the -mno-unaligned-access isn't present in `multilib.json` as we don't need to select a variant based on it.

vhscampos:
> LGTM. I guess that the -mno-unaligned-access isn't present in `multilib.json` as we don't need to select a variant based on it.

It's not present in order to select these new variants with or without `-mno-unaligned-access` in the command-line.

PR #590, opened 2024-12-09
Remaster picolibc patches against commit e3a4f137d85c5e9

statham-arm:
This deletes most of the previous patch 0002 (add bootcode for AArch64 FVPs), because it's now upstream in picolibc itself. The remaining vestige, defining picocrt_machines in the AArch32 picolibc build configuration, remains necessary until we conditionalise our definition of `test-machine` in our own builds.


ostannard (approved):
LGTM

PR #593, opened 2024-12-10
Fix delimiter so CMake reads string as a list

dcandler:
The libraries_supported property is a comma separated list, but CMake uses semi-colons to understand lists. Replacing the character will allow CMake to check which libraries are set within.


statham-arm (approved):
Ah, good. I feared you might have changed the JSON strings to use `;`, which would be ugly! I prefer this way round.

(I suppose another option would be to use an actual JSON _list_, but I assume that's 10 lines of cmakery instead of just 1?)

dcandler:
I did consider using an actual list in the JSON, but yes, it would be even more CMake scripting to find (and then correctly decrement) the length of the list, and leaving it comma separated is still just as readable.

PR #595, opened 2024-12-11
Add patch to fix volatile access bug

ostannard:
We have a customer affected by this bug, which was fixed after LLVM 19, and we'd like to do a release of LLVM-ET 19 with the fix.

LLVM PR: https://github.com/llvm/llvm-project/pull/111679

Original commit message:

[DAGISel] Keep flags when converting FP load/store to integer (#111679)

This DAG combine replaces a floating-point load/store pair which has no other uses with an integer one, but did not copy the memory operand flags to the new instructions, resulting in it dropping the volatile flag. This optimisation is still valid if one or both of the instructions is volatile, so we can copy over the whole MachineMemOperand to generate volatile integer loads and stores where needed.


patches/llvm-project/0001-libc-tests-with-picolibc-xfail-two-remaining-tests.patch
```diff
@@ -1,7 +1,7 @@
-From cb6cfe0e5fc29d3867f523900f0d46a4ef5dff1b Mon Sep 17 00:00:00 2001
+From 9fb80381b839fdbe15ddfa732bed5c25d6e4a5a2 Mon Sep 17 00:00:00 2001
 From: =?UTF-8?q?Dominik=20W=C3=B3jt?= <dominik.wojt@arm.com>
 Date: Mon, 16 Oct 2023 11:35:48 +0200
-Subject: [libc++] tests with picolibc: xfail two remaining tests
+Subject: [PATCH 1/9] [libc++] tests with picolibc: xfail two remaining tests
```
statham-arm, line 4: You should remaster the patch series using `format-patch -k`, to avoid getting these `[PATCH 1/9]` prefixes. Goes with the `git am -k` we advise for reapplying them.

(If you didn't do either, then `git am` without `-k` would also discard square-bracketed prefixes you actually wanted to keep, which there are lots of in LLVM.)

statham-arm (commented):
I know you're in a hurry, but could you please at least put _something_ in the commit message that allows people to find out what the patch is about without having to go and read the message in the subsidiary patch? "#111679" by itself means someone has to manually make up a URL to go and look that up in the LLVM project (and has to work out from context that it _is_ the LLVM project, because the PR number is too big to be picolibc).

A full URL to the PR would be useful. And maybe a one-sentence summary of what the fix is about.

statham-arm (approved):

Issue #555, opened 2024-11-05
std::unique_lock is not available


nolange:
The classes are very useful without threads (eq. using Irq locks).

Upstream has fixed this, but it didnt make it to version 19. Please cherry-pick those commits:

https://github.com/llvm/llvm-project/commit/6192f458f4fd2ca4e6f01515547034f89325e73c
https://github.com/llvm/llvm-project/commit/e9d5842ac7bd8d9b5700dc415910425ca796f4c

nolange:
Well, seems like `std::lock_guard` *did* end up in 19, but `std::unique_lock` did not. means only https://github.com/llvm/llvm-project/commit/e9d5842ac7bd8d9b5700dc415910425ca796f4c is necessary.

nolange:
No, its just in `main` for now, as far as I can see.

But you can get the patch from github from the commit: https://github.com/llvm/llvm-project/commit/e9d5842ac7bd8d9b5700dc415910425ca796f4c.patch ?

nolange:
fixed by #601

PR #601, opened 2024-12-17
Make std::unique_lock available

nolange:
Add commit e9d5842ac7bd from upstream as patch.

Closes: #555


voltur01:
Hi, thanks for the PR, it should be applied to the `llvm-19` branch though https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/tree/llvm-19 not `main`. Please update.

voltur01:
Thank you for the update!

Just realized we still have the list of patches in the CMake file in `llvm-19` branch, so this patch file name needs to be added here https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/blob/8abd9ca75d2506ed81afbac00b5658086eed967c/CMakeLists.txt#L291 to be applied.

voltur01:
This should work now, thank you! Would you mind updating the patch file number to 0011 since 0010 landed earlier today https://github.com/ARM-software/LLVM-embedded-toolchain-for-Arm/blob/llvm-19/patches/llvm-project/0010-Add-64bit-atomic-check-in-the-is_always_lock_free_pa.patch - I will approve and merge then. Thanks!

voltur01 (approved):

PR #603, opened 2024-12-17
Update the names of patch files to align with trunk

vrukesh:
Update the names of patch files to align with trunk.


dcandler:
I don't understand why this says it is aligning the patch files with trunk: the main branch has a different set of patches to the llvm-19 branch, so this isn't using the names from trunk.

The actual change to the CMake script looks correct since the names there match what is in the patch folder, it's only the commit message that is confusing.

voltur01 (approved):
LGTM

PR #605, opened 2024-12-18
Add AArch32 base variants with strict alignment

vhscampos:
This patch adds AArch32 base variants with `-mno-unaligned-access`.

Base variants are the set of libraries that provide the most compatible
functionality. As a consequence, they are built with strict alignment in
order to work nicely with client code that is compiled with or without
strict alignment.


pratlucas (approved):
LGTM.

nolange:
> As a consequence, they are built with strict alignment in order to work nicely with client code that is compiled with or without strict alignment.

I doubt that's the case, the code should work nicely without strict alignment, unless you explicitly disable unaligned load/store via some special register (Can't think of many reasons why you would).

The only observable effect would be [some slower routines](https://github.com/search?q=repo%3Abminor%2Fnewlib+__ARM_FEATURE_UNALIGNED&type=code).

can you describe some usecases where strict alignment  is necessary (on cores that don't use it by default)?
AFAIU Code should be compatible anyway, whatever your client code does unless you disable some bits in CP15.

in the cases where you absolutely need it, you could replace the few problematic functions or link the libraries from a "lower" arch like ARMv6.
That would be better than hurting everything by default.

vhscampos:
We have some internal customers that rely on strict alignment.

Bear in mind this is not the final picture. As this is an incremental work, in the near future we will add library variants with more optimised code generation, including unaligned access.

These base variants are a fallback option when no better option is available.

nolange:
> These base variants are a **fallback option** when no better option is available.

Maybe i misunderstand something, this patch builds **every armv7 target** with strict alignment?

My expectation when using multilib is that code is explicitly optimized for each arch.

PR #622, opened 2025-01-07
Add tests for AArch64 A-profile big endian library selection

vhscampos:
The tests check if the same variant is chosen with or without `-mno-unaligned-access`. For AArch64 A-profile big endian we provide only strictly aligned libraries.


simpal01 (approved):
LGTM

PR #623, opened 2025-01-08
Extend OmaxLTO options

nasherm:
This patch modifies extends OmaxLTO to make use of extra optimisation flags which affect LTO.

### DIFF
--- a/OmaxLTO.cfg
+++ b/OmaxLTO.cfg
@@ -1,5 +1,12 @@
 -flto=full \
+-fuse-linker-plugin \
 -fvirtual-function-elimination \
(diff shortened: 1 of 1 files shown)


nasherm:
> Any particular reason to not put -fuse-linker-plugin in its own line?

No reason, can easily move to it's own line

VladiKrapp-Arm:
> > Any particular reason to not put -fuse-linker-plugin in its own line?
> 
> No reason, can easily move to it's own line

For git blame friendliness reasons, if nothing else (:

stuij (approved):
LGTM!

PR #625, opened 2025-01-10
Fix issues with lit-exec-fvp.py

vhscampos:
This patch gives execution permissions to this Python script and fixes one wrong variable name.

Since this script is used as an executable in our lit tests, it ought to have execution rights.


ostannard (approved):
LGTM

PR #629, opened 2025-01-22
llvm-libc build: remove all references to libc-hdrgen.

statham-arm:
The 'newhdrgen' tool replaced it some time ago, and that's written in Python, so it doesn't need to be built separately with host tools at all.

As of upstream commit e0ae7793fca0c78, the migration is complete, and the old C++ libc-hdrgen doesn't exist at all any more. So we need to stop trying to build it – it will break our builds when no such build target exists.


pratlucas (approved):
LGTM.

PR #632, opened 2025-01-27
llvm-libc build: fix build failure with C++ libs disabled

statham-arm:
Testing the llvm-libc build after the previous commit, I found that the arm-multilib cmake script unconditionally tries to invoke build targets like `cxxlibs-configure` or `cxxlibs-build` in the individual library variant builds. Those targets don't exist in an llvm-libc build, because those set `"ENABLE_CXX_LIBS": "OFF"` in their per-variant JSON files.

This commit applies the simplest possible fix: _make_ targets of those names in `arm-runtimes`, even if nothing is actually in them. Then invoking them from the higher-level build script is a NOP instead of a failure.

### DIFF
--- a/arm-runtimes/CMakeLists.txt
+++ b/arm-runtimes/CMakeLists.txt
@@ -771,6 +771,14 @@ if(ENABLE_CXX_LIBS)
         endforeach()
     endif()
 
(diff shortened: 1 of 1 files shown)


dcandler (approved):
LGTM

PR #636, opened 2025-02-17
Fix performance patch no longer applying due to upstream change

stuij:
Change in question: #126168


voltur01 (approved):

walkerkd:
Change looks good to me.

PR #639, opened 2025-02-25
Disable debug symbols in picolibc builds

voltur01:


smithp35 (approved):
I'm comfortable with turning off debug for the libraries build. It is the only option that is documented to work with both a buildtype of minsize and release.
There's a possibility that we could detect if the CMake build type is debug and only put this in if that's the case, but I expect debugging the tools to be a separate use case from debugging the libraries.

PR #642, opened 2025-05-08
Remove unused copies of `meson-cross-build.txt.in`/`to_meson_list.cmake`

s-barannikov:
The used copies are in `arm-runtimes` directory.


s-barannikov:
`to_meson_list.cmake` is also duplicated, let me know if I should delete its unused copy too.

voltur01:
Hi, I can confirm that the files were moved and should be deleted in the old place - please add the second file and I will approve.

Note that this project is not actively maintained anymore so I created similar PR for the Arm Toolchain project https://github.com/arm/arm-toolchain/pull/322

Do you have any specific reason to prefer this project to Arm Toolchain?

s-barannikov:
Hi,

No specific reason, I just didn't know about the other project (haven't opened readme for a while). Thanks for pointing that out!

voltur01 (approved):
Thank you for the update!

